PR #5, opened 2019-06-21
SpaceDamage serialization and sdl.surface caching.

Lemonymous:
These functions might not be placed where you want them, so let me know if you want them moved elsewhere.

I am unsure of your naming convention of globals.
The global for max surfaces before evicting:
sdlext.surface_cache_max


kartoFlane:
Naming convention for globals is, unfortunately, all over the place, as the original mod loader and the game named things differently. Even the game code itself is not 100% consistent in this regard. Generally the convention depends on which section I'm adding to:
- global tables or functions pertaining to mod loader / things that are generally supposed to be part of modding API: `PascalCase`
- local helper functions and variables: `camelCase`
- utility global functions that are kinda related to game's own global functions, or are kind of low-level / have a narrow scope of responsibility: `lower_snake_case`

Overall, naming could definitely use some cleaning up. 

The serialization/deserialization section of `misc.lua` is starting to get big enough that it might warrant getting its own file the next time we add something to it.

scripts/mod_loader/altered/misc.lua
```diff
@@ -231,3 +235,52 @@ PointList.GetLuaString = function(self)
 	return string.format("CreatePointList(%s)", save_table(extract_table(self)))
 end
 PointList.GetString = PointList.GetLuaString
+
+local spaceDamage = {
```
kartoFlane:
```suggestion
local spaceDamageKeys = {
```

scripts/mod_loader/altered/misc.lua
```diff
@@ -231,3 +235,52 @@ PointList.GetLuaString = function(self)
...
+	"sItem",
+	"sPawn",
+	"sScript",
+	"sSound"
+}
+
+function SerializeSpaceDamage(self)
+	local result = {}
+	
+	for _, i in ipairs(spaceDamage) do
```
kartoFlane:
```suggestion
	for _, i in ipairs(spaceDamageKeys) do
```

scripts/mod_loader/altered/misc.lua
```diff
@@ -231,3 +235,52 @@ PointList.GetLuaString = function(self)
...
+		result[i] = self[i]
+	end
+	
+	return result
+end
+
+function DeserializeSpaceDamage(self)
```
kartoFlane: To follow the naming convention of the other functions:
```suggestion
function CreateSpaceDamage(self)
```

scripts/mod_loader/altered/misc.lua
```diff
@@ -231,3 +235,52 @@ PointList.GetLuaString = function(self)
...
+	
+	for i, v in pairs(self) do
+		result[i] = v
+	end
+	
+	return result
+end
+
+SpaceDamage.GetLuaString = function(self)
+    return string.format("DeserializeSpaceDamage(%s)", save_table(SerializeSpaceDamage(self)))
```
kartoFlane:
```suggestion
    return string.format("CreateSpaceDamage(%s)", save_table(SerializeSpaceDamage(self)))
```

scripts/mod_loader/altered/misc.lua
```diff
@@ -231,3 +235,52 @@ PointList.GetLuaString = function(self)
...
+	"loc",
+	"sAnimation",
+	"sImageMark",
+	"sItem",
+	"sPawn",
+	"sScript",
+	"sSound"
+}
+
+function SerializeSpaceDamage(self)
```
kartoFlane: "Serialization" typically refers to converting an object to a string/byte format that can be saved to a file. Here we're converting to a table, so the name doesn't quite fit. Something like `SpaceDamageToTable`  might be more descriptive

Lemonymous: Right, I suppose the name would have made more sense if the save_table call was inside the function.

PR #8, opened 2019-06-23
Fixed Board.IsMissionBoard and added Board.IsTipImage

Lemonymous:
Board type does not contain an operator for comparing one instance to another.

### DIFF
--- a/scripts/mod_loader/altered/missions.lua
+++ b/scripts/mod_loader/altered/missions.lua
@@ -56,6 +56,7 @@ end
 
 local oldBaseUpdate = Mission.BaseUpdate
 function Mission:BaseUpdate()
+	Board.isMission = true
 	modApi.current_mission = self
 	modApi:processRunLaterQueue(self)
 
@@ -196,6 +197,7 @@ function Mission:BaseStart(suppressHooks)
 		end
 	end
 
+	Board.isMission = true
 	self.Board = Board
 	self.QueuedSpawns = {}
 
--- a/scripts/mod_loader/altered/missions.lua
+++ b/scripts/mod_loader/altered/missions.lua
@@ -56,8 +56,7 @@ end
 
 local oldBaseUpdate = Mission.BaseUpdate
 function Mission:BaseUpdate()
-	Board.isMission = true
-	modApi.current_mission = self
+	Board.mission = self
 	modApi:processRunLaterQueue(self)
 
 	oldBaseUpdate(self)
@@ -74,7 +73,7 @@ end
 
 local oldBaseDeployment = Mission.BaseDeployment
 function Mission:BaseDeployment()
-	modApi.current_mission = self
+	Board.mission = self
 	oldBaseDeployment(self)
 	self.Board = Board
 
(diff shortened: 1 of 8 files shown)


scripts/mod_loader/altered/missions.lua
```diff
@@ -56,6 +56,7 @@ end
 
 local oldBaseUpdate = Mission.BaseUpdate
 function Mission:BaseUpdate()
+	Board.isMission = true
 	modApi.current_mission = self
```
kartoFlane: Could *maybe* kill two birds with one stone here, and store reference to the mission inside of Board, like so:
```lua
Board.mission = self
```
This would allow us to remove `modApi.current_mission`, so we'd have less code to maintain. It would also be a little easier to reason about, since we can say that as soon as we have a reference to `Board`, we also have a reference to current mission.

Lemonymous: It's definitely an interesting thought. The change in this pull request sets `Board.isMission = true` in `BaseStart` while `current_mission = true` in `BaseDeployment`, so there is some separation between the two.

Since `Board` is available when previewing missions, `Board.mission` will have to be set in `BaseDeployment` instead. The net result will be that `GetCurrentMission` will not work in hooks originating from `BaseStart`. A small price to pay maybe.

Lemonymous: Actually. I did some more testing. Turns out the `Board` that you get while viewing a tipimage is the one inside the tipimage. So by not having the separation, GetCurrentMission() will return nothing while viewing tipimages, even inside a mission.

kartoFlane: I think having `Board.mission` available in `BaseStart` would be preferable (or maybe even earlier, in `Mission:Initialize`), so that we don't limit potential use cases. If I remember, the `Board` global is correctly set when the mission is being created, so it shouldn't be a problem.

Random thought: I wonder if it's possible to only set the field once, so that we don't have to re-set it every game update. Maybe adding a function to `BoardClass`, like `BoardClass:GetMission()`, that does nothing by default, and then overriding this function on each `Board` instance to return its associated mission via upvalue will get it to stick... Something like:
```lua
-- board.lua
BoardClass.GetMission = function()
end

-- missions.lua
function Mission:Initialize()
    -- ...
    Board.GetMission = function()
        return self
    end
end
```
Though if my understanding of why the unsetting happens is correct, it's not going to work.

Comparing the same instance of the board to itself *does* return true (which is why assumed it worked for other boards as well), so maybe we can abuse that to keep a table of all boards, indexed by their associated mission, and then iterate it, performing the comparison inside of a `pcall()` for error handling.

Regarding tip images, overall I'm in favor of them not being aware of current mission. It will limit use cases a little, but tip images should always be the same, regardless of current mission, and this will act as a soft enforcement of this rule. If they really need to, modders can always store a reference to current mission in a variable.

kartoFlane:
You can revert the last two commits on this branch if they're not intended to be included in this PR.

PR #14, opened 2019-07-23
Fixed HangarGetSelectedMechs

Lemonymous:
With the addition of PawnTable in modapi/pawn.lua; every table deriving from it will have several GetImage overrides.
This change ensures only one call is made to table.insert(fetchedMechs... for each mech. (in modui/hangar.lua)

### DIFF
--- a/scripts/mod_loader/modui/hangar.lua
+++ b/scripts/mod_loader/modui/hangar.lua
@@ -130,9 +130,10 @@ end
 
 local function overrideGetImages()
 	pawns = {}
-
+	local wasPrimaryCallExecuted = false
+	
 	for k, v in pairs(_G) do
-		if type(v) == "table" and v.Health and v.Image then
+		if type(v) == "table" and v ~= PawnTable and v.Health and v.Image then
 			table.insert(pawns, k)
 
 			if v.GetImage then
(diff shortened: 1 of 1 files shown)


kartoFlane:
Good catch. I would also add an exclusion of `PawnTable` to the filtering logic a few lines above, so that we don't override the base table:
```lua
if type(v) == "table" and v ~= PawnTable and v.Health and v.Image then
```

Though now that I look at it, even with this fix, this function will break if there's a pawn that refers to another pawn class' `GetImage` function directly, like so:
```lua
SomeMech.GetImage = function(self)
    return PunchMech.GetImage(self)
end
```
Hardly common, but it *could* happen. Since at that point in time, `GetImage` is still the overridden function, not the original one, the additional argument will not be passed, and `PunchMech` will also be added to the list of fetched mechs.

An approach like this should fix this:
```lua
v.GetImage = function(self)
    local isPrimaryCall = not wasPrimaryCallExecuted

    if
        isPrimaryCall             and
        IsHangarWindowlessState() and
        #fetchedMechs < 3
    then
        table.insert(fetchedMechs, k)
    end

    if isPrimaryCall then
        wasPrimaryCallExecuted = true
    end
    local result = oldGetImages[k](self)
    if isPrimaryCall then
        wasPrimaryCallExecuted = false
    end

    return result
end
```
The `wasPrimaryCallExecuted` local would need to be defined outside the `for` loop at line 134

Lemonymous:
That's also a good catch, and a nice solution.

PR #15, opened 2019-07-23
Generalized custom difficulty compatibility

Lemonymous:
Removed full overrides of vanilla functions.
Swapping out GetDifficulty with GetRealDifficulty before calling functions instead.

### DIFF
--- a/scripts/mod_loader/altered/difficulty.lua
+++ b/scripts/mod_loader/altered/difficulty.lua
@@ -16,193 +16,45 @@ end
 
 -- /////////////////////////////////////////////////////////////////////////////////////
 -- Tweak existing code to work with custom difficulty levels
--- Replacing instances of GetDifficulty() with GetBaselineDifficulty()
+-- Replacing instances of GetDifficulty() with GetRealDifficulty()
 
-function Mission_Final:StartMission()
-	self:GetSpawner():SetSpawnIsland(5)
-	local pylons = extract_table(Board:GetZone("pylons"))
-	for i,v in ipairs(pylons) do
-		Board:BlockSpawn(v,BLOCKED_PERM)
-	end
-	
-	if GetBaselineDifficulty() == DIFF_HARD then
-		Board:SpawnPawn(random_element(self.BossList))
-	end
-end
+local functions = {
+	{"Mission_Final", "StartMission"},
+	{"getEnvironmentChance"},
+	{"Mission_SpiderBoss", "SpawnSpiderlings"},
+	{"Mission", "GetKillBonus"},
+	{"Mission", "GetStartingPawns"},
+	{"Mission", "GetSpawnsPerTurn"},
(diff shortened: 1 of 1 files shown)


kartoFlane:
Definitely a neat solution, and much better than the previous approach which involved a lot of ugly code duplication. Though the part that extracts data from the table could use some clearer variables names or maybe comments, it took me a moment to figure out what exactly it was doing. Clear variable names are very valuable in such cases, even if they have to be long to convey their meaning.

Lemonymous:
I'm having some trouble giving the variables in the for loops good names.

How about something like this? Suggestions are welcome.

```lua
for _, keyList in ipairs(functionList) do
	local tbl = _G
	local finalKey = nil
	
	-- search through keyList until we find the table and key,
	-- so that tbl[key] is the function at the end of the sequence:
	-- {table, sub_table, .. , function}
	for _, key in ipairs(keyList) do
		if type(tbl[key]) == 'function' then
			finalKey = key
			-- function found; end search.
			break
			
		elseif type(tbl[key]) == 'table' then
			tbl = tbl[key]
			-- table found; step into table, and continue search.
			
		else
			-- no function was found; end search.
			break
		end
	end```

**Comment by kartoFlane** [comment]
I'd suggest a slightly different approach, since upon closer inspection, the table-searching functionality does not appear to be required (and it was the most confusing part). So we can simply associate a table name with a function list, and iterate that. Separating the table names from the function names also gives it more clarity, both in the source data table, and in code that processes it, since we don't have to manually extract the last key.
```lua
local functionsToReplace = {
    "_G" = { "getEnvironmentChance" },
    "Mission_Final" = { "StartMission" },
    "Mission_SpiderBoss" = { "SpawnSpiderlings" },
    "Mission" = {
        "GetKillBonus",
        "GetStartingPawns",
        "GetSpawnsPerTurn",
        "GetMaxEnemy",
        "GetSpawnCount"
    }
}

local function buildReplacementFunction(sourceFn)
    return function(...)
        local oldGetDiff = GetDifficulty

        -- Only replace if we need to, to account for
        -- nested function calls
        if GetDifficulty ~= GetRealDifficulty then
            GetDifficulty = GetRealDifficulty
        end

        local result = sourceFn(...)

        if GetDifficulty == GetRealDifficulty then
            GetDifficulty = oldGetDiff
        end

        return result
    end
end

for tableName, functionList in pairs(functionsToReplace) do
    local tbl = _G[tableName]

    for _, functionName in ipairs(functionList) do
        local fn = tbl[functionName]
        if type(fn) == "function" then
            tbl[functionName] = buildReplacementFunction(fn)
        end
    end
end
```
If the table-searching functionality is needed, then it can be extracted to a function like `findTableByPath` or `findFirstOccurenceOfFunctionInTablePath` (a bit on the extreme end, but it does adequately describe what the function does)

PR #21, opened 2020-02-15
Truncated PilotList in hangar

Lemonymous:
Truncates pilot list while in the hangar to the 13 first selected, while still allowing all pilots to be found from pods and perfect island rewards.

### DIFF
--- a/scripts/mod_loader/modapi/global.lua
+++ b/scripts/mod_loader/modapi/global.lua
@@ -16,13 +16,18 @@ function CreatePilotPersonality(label, name)
 	return t
 end
 
+PilotListExtended = shallow_copy(PilotList)
+
 function CreatePilot(data)
 	_G[data.Id] = Pilot:new(data)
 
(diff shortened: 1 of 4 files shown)


scripts/mod_loader/modapi/global.lua
```diff
@@ -16,13 +16,18 @@ function CreatePilotPersonality(label, name)
...
 end
 
+PilotListExtended = shallow_copy(PilotList)
+
 function CreatePilot(data)
 	_G[data.Id] = Pilot:new(data)
 
 	-- Make sure we don't create duplicates if the PilotList
 	-- already contains entry for this pilot
 	if data.Rarity ~= 0 and not list_contains(PilotList, data.Id) then
```
KnightMiner: Should this be checking PilotListExtended? Otherwise you may get duplicates after 13

Lemonymous: Yeah, you're right. It should.

PR #24, opened 2020-02-16
Pilot Arrange UI - add background "hangar" slot image

tos-x:


scripts/mod_loader/modui/pilot_arrange.lua
```diff
@@ -111,13 +111,30 @@ local function createUi()
...
+		
 		local function addPilotButton(i, pilotId)
 			local pilot = _G[pilotId]
 			local col = (i - 1) % portraitsPerRow
 			local row = math.floor((i - 1) / portraitsPerRow)
 			
 			local surface = sdl.scaled(2, sdlext.surface("img/portraits/pilots/"..pilotId..".png"))
-			local button = Ui()
+			local button
+			button = Ui()
```
kartoFlane: There's no need to split this over two lines

scripts/mod_loader/modui/pilot_arrange.lua
```diff
@@ -111,13 +111,30 @@ local function createUi()
 			end
 		end
 		
+		local function addHangarButton(i)
+			local col = (i - 1) % portraitsPerRow
+			local row = math.floor((i - 1) / portraitsPerRow)
+			
+			local surface = sdlext.surface("resources/mods/ui/pilot_arrange_hangar.png")
```
kartoFlane: This creates a surface each time a button is created, which is very inefficient (particularly because this is an actual SDL object). It's better to move this to the top of the file next to `MAX_PILOTS`, so that the surface is only created once, and then re-used by UI objects.

scripts/mod_loader/modui/pilot_arrange.lua
```diff
@@ -127,6 +144,14 @@ local function createUi()
...
 				})
 				:addTo(scrollarea)
+				--[[
+			if i <14 then				
+				button.decorations[1].color = sdl.rgb(13, 15, 23)
+				button.decorations[1].hlcolor = hlcolor or deco.colors.buttonhl
+				button.decorations[1].bordercolor = sdl.rgb(35, 42, 59)
+				button.decorations[1].borderhlcolor = borderhlcolor or deco.colors.buttonborderhl
+			end
+			--]]
```
kartoFlane: Typically you don't include commented-out code in pull requests ;)

PR #26, opened 2020-02-16
Add dll which adds multiple new features via memory manipulation

Lemonymous:
In an effort to not have a too large pull request at a later date, I finished the functions I had as much as I could, and open this pull request now.

No need to rush to accept it. Check it out, and see if it holds to your standards.


Lemonymous:
Added documentation for Board here to help with the review process.
https://github.com/Lemonymous/ITB-ModLoader/wiki/Board

Lemonymous:
Documentation for Pawn functions:
https://github.com/Lemonymous/ITB-ModLoader/wiki/Pawn

scripts/mod_loader/modapi/pawn.lua
```diff
@@ -205,39 +234,398 @@ BoardPawn.ApplyDamage = function(self, spaceDamage)
...
+BoardPawn.SetMoveSkill = function(self, skill)
+	CUtils.PawnSetMoveSkill(self, skill)
+end
+
+BoardPawn.SetRepairSkill = function(self, skill)
+	CUtils.PawnSetRepairSkill(self, skill)
+end
+
+
+BoardPawn.GetPilotId = function(self)
```
kartoFlane: Seems like there's a lot of functions related to pilots here. Is there no way to separate them from pawn entity, and make them work with pilots in the player's inventory?

scripts/mod_loader/modapi/board.lua
```diff
@@ -1,96 +1,696 @@
...
+end
+
+local board_metatable
+local doInit = true
+local oldSetBoard = SetBoard
+function SetBoard(board)
+	if board ~= nil then
+		
+		if doInit then
+			doInit = nil
```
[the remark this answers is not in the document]
Lemonymous: `BoardClass` as well as `board_metatable` is set up only once ever.
For every subsequent GetBoard call we can simply set the metatable of board to `board_metatable`

scripts/mod_loader/modapi/board.lua
```diff
@@ -1,96 +1,696 @@
...
-	Tests.AssertEquals("userdata", type(self), "Argument #0")
+	Tests.AssertSignature{
+		ret = "boolean",
+		func = "IsTipImage",
+		params = { self },
+		{ "userdata|GameBoard&" }
+	}
 	
-	return self.isMission == nil
+	return isTipImage(self)
```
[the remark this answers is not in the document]
Lemonymous: I don't have any strong reason for one or the other, so you can choose.

scripts/mod_loader/modapi/pawn.lua
Lemonymous: Well, we don't have any real `Pilot` object, so it has to come from a pawn at some point. I suppose we could have a fake `Pilot` object that we could get with `Pawn:GetPilot()`, which would hold all pilot functions.
I don't think we should just expect 3 pilots from the main 3 mechs by the way, since that limitation might be lifted in the future.

scripts/mod_loader/modapi/board.lua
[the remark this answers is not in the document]
kartoFlane: Well, in this case the local function seems pretty pointless, so it should be removed :)

scripts/mod_loader/modapi/board.lua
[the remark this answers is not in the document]
kartoFlane: Ah, I missed that the `SetUserdataMetatable` call was outside the if block

scripts/mod_loader/modapi/pawn.lua
kartoFlane: I thought about the same thing regarding a fake `Pilot` object, though in that case the question is, how are the pilot-modifying functions implemented in C++ side?  
If the `Pilot` object has to proxy through a pawn memory address, then it's problematic, since mods could hold on to the `Pilot` object reference, while the player assigns a different pilot to the mech. The `Pilot` object would still issue calls through the pawn, modyfing another pilot's data.

A quick glance at the C++ code seems to suggest there's a separate `PilotData` structure, so it seems like it could be feasible?

The vanilla pawn functions relating to the pilot are few in number, and only *return* data. With the number of new functions added here, as well as the fact that they can be used to modify data, I think it's better to separate them. It also makes it easier to manage in the future, since the `Pilot` object and the new functions can be placed in a `pilot.lua` file.

Lemonymous: I'll do some research and see if I can separate Pilot from Pawn properly. It does appear pilots exists entirely separate from pawns, so it should be doable. If it works out I think it will be fairly easy to move pilots inbetween mechs, and hopefully create new pilot instances at will as well.

scripts/mod_loader/__scripts.lua
```diff
@@ -1,7 +1,7 @@
 local scripts = {
 	"hard-alter",
-	"Cutils",
 	
+	"cutils",
```
KnightMiner: I think Git is not a particular fan of renaming that only changes capitalization. I have seen people do it before by deleting or a more significant rename, then renaming it back in 2 commits.

Lemonymous: Hm.. So I see. I tried to change the file to all lower case to match the other files there, but it doesn't look like git allowed that.
Anyways; having trimmed the loading of `cutils.dll` down to a single line of code, maybe it could simply go into another file instead.

scripts/mod_loader/modapi/mechedit.lua
```diff
@@ -0,0 +1,509 @@
...
+		params = { highlightable },
+		{ "boolean|bool" }
+	}
+	
+	if not Game then
+		return
+	end
+	
+	CUtils.SetMechEditNameHighlightable(highlightable)
+end
```
kartoFlane: ⚠ Doesn't appear to be working in my version of the game - the function completes, but the mech's name is still highlightable / editable

scripts/mod_loader/modapi/mechedit.lua
```diff
@@ -0,0 +1,509 @@
...
+		params = { highlightable },
+		{ "boolean|bool" }
+	}
+	
+	if not Game then
+		return
+	end
+	
+	CUtils.SetMechEditPilotNameHighlightable(highlightable)
+end
```
kartoFlane: ⚠ Doesn't appear to be working in my version of the game - the function completes, but the pilot's name is still highlightable / editable

scripts/mod_loader/modapi/mechedit.lua
```diff
@@ -0,0 +1,509 @@
...
+		params = { highlightable },
+		{ "boolean|bool" }
+	}
+	
+	if not Game then
+		return
+	end
+	
+	CUtils.SetMechEditInstallHighlightable(highlightable)
+end
```
kartoFlane: ❗ Attempting to call this causes an error; it appears `CUtils.SetMechEditInstallHighlightable` is missing?

scripts/mod_loader/modapi/mechedit.lua
```diff
@@ -0,0 +1,509 @@
...
+
+-- Hovered
+function mechedit.IsTestHovered()
+	if not Game then
+		return false
+	end
+	
+	return CUtils.IsMechEditTestHovered()
+end
```
kartoFlane: ℹ Continues returning `true` while in test mech scenario, probably shouldn't?

scripts/mod_loader/modapi/board.lua
```diff
@@ -1,96 +1,691 @@
...
+BoardClass.IsForestFire = function(self, loc)
+	Tests.AssertSignature{
+		ret = "bool",
+		func = "IsForestFire",
+		params = { self, loc },
+		{ "userdata|GameBoard&", "userdata|Point" }
+	}
+	
+	return CUtils.IsForestFire(self, loc)
 end
```
kartoFlane: ❗ Calling this causes an error; it appears that `CUtils.IsForestFire` is missing

scripts/mod_loader/modapi/board.lua
```diff
@@ -1,96 +1,691 @@
...
+BoardClass.IsForest = function(self, loc)
+	Tests.AssertSignature{
+		ret = "bool",
+		func = "IsForest",
+		params = { self, loc },
+		{ "userdata|GameBoard&", "userdata|Point" }
+	}
+	
+	return CUtils.IsTileForest(self, loc)
+end
```
kartoFlane: ⚠ Always returns `false` in my version of the game

scripts/mod_loader/modapi/board.lua
```diff
@@ -1,96 +1,691 @@
...
+		else
+			self:SetTerrain(loc, TERRAIN_WATER)
+		end
+		
+		self:SetLava(loc, false)
+		self:SetAcid(loc, false)
+	elseif self:GetTerrain(loc) == TERRAIN_WATER then
+		self:SetTerrain(loc, TERRAIN_ROAD)
+	end
+end
```
kartoFlane: ℹ The sinking animation looks kind of awkward, since the tile changes to water first, and then starts sinking. Probably not possible to fix?

scripts/mod_loader/modapi/board.lua
```diff
@@ -1,96 +1,691 @@
...
+			hp_max = 1
+		else
+			hp_max = math.max(1, math.min(4, hp_max))
+		end
+		
+		local hp = CUtils.GetTileHealth(self, loc.x, loc.y)
+		
+		self:SetBuilding(loc, hp, hp_max)
+	end
+end
```
kartoFlane: ❗ Calling this causes an error on line `local hp = CUtils.GetTileHealth(self, loc.x, loc.y)`, the function expects a `Point`

scripts/mod_loader/modapi/board.lua
```diff
@@ -1,96 +1,691 @@
...
+BoardClass.MarkGridLoss = function(self, loc, grid_loss)
+	Tests.AssertSignature{
+		ret = "void",
+		func = "MarkGridLoss",
+		params = { self, loc, grid_loss },
+		{ "userdata|GameBoard&", "userdata|Point", "number|int" }
+	}
+	
+	CUtils.TileMarkGridLoss(self, loc, grid_loss)
 end
```
kartoFlane: ❗ Calling this causes an error: `bad argument #4 to 'TileMarkGridLoss' (number expected, got no value)`  
I called: `Board:MarkGridLoss(Board:GetHighlighted(), 2)`, happens both over regular tiles as well as buildings, in test mech scenario

scripts/mod_loader/modapi/game.lua
```diff
@@ -0,0 +1,177 @@
...
+	CUtils.SetGridDef(new)
+end
+
+GameClass.SubGridDef = function(grid_def)
+	GameClass.AddGridDef(-grid_def)
+end
+
+GameClass.GridBlink = function()
+	CUtils.DoGridBlink()
+end
```
kartoFlane: ❗ Fails with error `bad argument #1 to 'DoGridBlink' (number expected, got no value)`

scripts/mod_loader/modapi/mechedit.lua
```diff
@@ -0,0 +1,509 @@
...
+		params = { highlightable },
+		{ "boolean|bool" }
+	}
+	
+	if not Game then
+		return
+	end
+	
+	CUtils.SetMechEditHealthIconHighlightable(highlightable)
+end
```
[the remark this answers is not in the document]
Lemonymous: The `Highlightable` property of the ui elements come in three variations.
1. The game does not constantly set it. These we can toggle with fire and forget functions. (example: `MechEdit.SetMoveHighlightable`)
2. The game constantly sets it -> then we get to act -> then the game updates. We can set this each frame to override the default behavior. (example: `MechEdit.SetNameHighlightable`)
3. The game constantly sets it -> then the game updates. We don't get to act before the game updates, so we cannot do anything in this case. (example: `MechEdit.SetHealthHighlightable`)
For the Health ui object in particular, I believe it is due to the game always checking if mafan is present or not, and setting the ui object accordingly.

scripts/mod_loader/modapi/mechedit.lua
Lemonymous: I have updated the documentation with `Set..Highlightable` functions with specifics on which functions need to be set each frame, and which are fire and forget functions.
In this case, `MechEdit.SetNameHighlightable` must be called each frame the vanialla behavior should be overridden.

scripts/mod_loader/modapi/mechedit.lua
Lemonymous: I have updated the documentation with `Set..Highlightable` functions with specifics on which functions need to be set each frame, and which are fire and forget functions.
In this case, `MechEdit.SetPilotNameHighlightable` must be called each frame the vanilla behavior should be overridden.

scripts/mod_loader/modapi/mechedit.lua
Lemonymous: Fixed in commit `c305de7`.
Note that this is another case where the function must be called each frame the vanilla behavior should be overridden.

scripts/mod_loader/modapi/mechedit.lua
Lemonymous: That is an oversight. I will look into this.

scripts/mod_loader/modapi/board.lua
Lemonymous: Nice catch. IsTileForest was incorrectly checking if there was a forest fire.

scripts/mod_loader/modapi/board.lua
Lemonymous: Fixed in `d3d8d13`

scripts/mod_loader/modapi/board.lua
Lemonymous: Fixed in `06dd4ac`
Note that this is another function that must to be called each frame grid loss should be marked on a building.

scripts/mod_loader/modapi/board.lua
Lemonymous: The function is quite hacky. I have found addresses that could do the sinking without swapping terrain like this. There are a lot of moving parts involved though, so it would have to be at some point in the future.
If this function is not good enough at present, I can remove it.

scripts/mod_loader/modapi/board.lua
Lemonymous: Fixed in `2927ad5`

scripts/mod_loader/modapi/board.lua
Lemonymous: I have looked at this again, and I actually cannot see what you mean is the awkward part. It functions exactly like the acid barrels do, but without the splash. For a mod that wants to have a sinking tile, I think this function should be sufficient to work with.

scripts/mod_loader/modapi/game.lua
Lemonymous: Grid blink appears to be more involved than it first appeared, so I have removed this function.

scripts/mod_loader/modapi/mechedit.lua
Lemonymous: Fixed in dc1389d.
All `MechEdit.Is[..]Hovered` functions will now always return `false` whenever we are not in the `Region` overview.

scripts/mod_loader/modapi/gamestate.lua
```diff
@@ -0,0 +1,30 @@
...
+	return CUtils.IsTransition()
+end
+
+function GameState.IsEnteringHangar()
+	return CUtils.IsHangar() and CUtils.IsTransition() and not CUtils.IsLeavingHangar()
+end
+
+function GameState.IsLeavingHangar()
+	return CUtils.IsLeavingHangar()
+end
```
Lemonymous: Documentation for functions here:
https://github.com/Lemonymous/ITB-ModLoader/wiki/GameState

scripts/mod_loader/modapi/board.lua
kartoFlane: I see, I guess I never paid too much attention to it /shrug

PR #27, opened 2020-02-17
Removes Pilots and Statistics errors.

Lemonymous:
Added initialization of all entries of pilots/mechs/weapons/enemies found in the profile data, in order to be able to remove a mod without having to suffer lingering errors on startup and statistics screen.


Lemonymous:
With 4 commits at current, I believe this is the final commit to add this feature. Barring any mistakes.

PR #28, opened 2020-02-21
Events

Lemonymous:
The first step in making the mod loader events driven.

Code from `altered/mission` has been moved to `altered/mission_additional` where it is run via events instead of inserted directly into the functions where the code needed to be executed. It should be functionally the same as the old code.

A next step could be to move code out of `altered/mission_additional` to group everything that logiocally relates to the same features together.

### DIFF
--- a/scripts/mod_loader/altered/missions.lua
+++ b/scripts/mod_loader/altered/missions.lua
@@ -1,18 +1,39 @@
 
-function Mission:OnSerializationStart(t)
-	t.MissionEndImpl = self.MissionEndImpl
-	t.MissionEnd = self.MissionEnd
-	t.Board = self.Board
+-- this file contains modifications to original functions
+-- as well as additional useful Mission functions.
 
-	self.MissionEndImpl = nil
-	self.MissionEnd = nil
-	self.Board = nil
+local function isSerializable(v)
+	local t = type(v)
+	
+	if t == "table" then
+		return true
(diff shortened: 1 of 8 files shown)


scripts/mod_loader/altered/missions.lua
```diff
@@ -160,134 +132,27 @@ function BuildIsBoardBusyPredicate(board)
...
-		modLoaderHooksFired = true
-	end
-	
-	return isDone
-end
-
-function Mission:IsEnvironmentEffect()
-	return true
-end
+-- Why have this function been changed to always return true?
```
kartoFlane: It allows for environment hooks to work in all missions (as far as I understand; this has been in the mod loader before I took over maintenance)

scripts/mod_loader/altered/missions.lua
```diff
@@ -160,134 +132,27 @@ function BuildIsBoardBusyPredicate(board)
...
+-- we call it manually when we detect the player leaving the test mech scenario.
+--[[function Mission_Test:MissionEnd()
+	-- DON'T call the default MissionEnd
+	-- Mission.MissionEnd(self)
 
-function Mission_Test:BaseStart()
-	Board.isMission = true
-	Mission.BaseStart(self, true)
+	modApi:fireTestMechExitedHooks(self)
+end]]
```
kartoFlane: Remove commented out code

PR #36, opened 2020-04-23
Added backwards compatibility code to GetText for modded weapon ugprades

Lemonymous:
With ITB version 1.2 - weapon upgrade names stored in `MyWeapon.UpgradeList` are ignored. This change reenables that functionality so old mods using that method don't need to be updated to work as they did before.

### DIFF
--- a/scripts/mod_loader/altered/text.lua
+++ b/scripts/mod_loader/altered/text.lua
@@ -43,3 +43,10 @@ function GetText(id, r1, r2, r3)
 		return text
 	end
 end
+
+local vanillaGetPilotDialog = GetPilotDialog
+function GetPilotDialog(personality, event)
+	local result = vanillaGetPilotDialog(personality, event)
+	
+	return GetPilotDialog_Deprecated(personality, event)
+end
--- a/scripts/mod_loader/altered/text.lua
+++ b/scripts/mod_loader/altered/text.lua
@@ -43,3 +43,14 @@ function GetText(id, r1, r2, r3)
 		return text
 	end
 end
+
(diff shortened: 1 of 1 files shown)


scripts/mod_loader/altered/text.lua
```diff
@@ -27,6 +27,26 @@ function GetText(id, r1, r2, r3)
...
+				end
+			end
+		end
+		
+		return text
 	end
 end
+
+local vanillaGetPilotDialog = GetPilotDialog
+function GetPilotDialog(personality, event)
```
KnightMiner, line 48: Was this accidentally committed? Looks a bit incomplete

Lemonymous: Thank you for pointing that out. Seems I commited the wrong code.

kartoFlane:
Looks good, though I think it'd be better to have this updated once at init, since `GetText()` is called every frame for every visible text (and some more, I think)... Little additions here can quickly snowball into noticeable delays.

We could probably hijack `Skill.new` to look into its argument and add `UpgradeList` entries into `Weapon_List`. Even changes to `UpgradeList` after the skill is created can be detected via metatables, though of course that solution will be more complex and less robust.

Lemonymous:
That sounds like premature over-optimization to me. Are you sure adding extra complexity and moving parts would be the best choice here?
For mods that don't use the deprecated methods this will add a single `if check` for upgrades and one more function call on the stack and an `if check` for dialogs.
Sure, it is something, but without testing the cost of it I would not opt for a more complex option personally.

PR #38, opened 2020-05-12
Add support for a weapon selector UI and mods adding weapons directly

KnightMiner:
This pull request adds in several features:

**`modApi` functions**
* `modApi:addWeapon(id, enabled)`: This function will add a new weapon to mod loader's weapon deck, which will be used in games. `enabled` is optional, if set will determine the weapons's default status. If unset defaults to the previous value, or true.
* `modApi:getWeaponDeck()`: Gets a list of all enabled weapons
* `modApi:isDefaultWeapon(id)`: Returns true if the weapon is a default vanilla weapon.

**Altered**
* `initializeDecks` and `getWeaponDrop` were modified to use the `modApi` weapon list.
* `Skill:GetRarity` is modified to allow setting `CustomRarity` to bypass vanilla logic for rarity.
* `Skill:GetUnlocked` is a function that returns when a weapon is unlocked. Weapons that are not unlocked will not be added to the weapon deck

**UI**
* Added the weapon deck selector UI
* This UI allows checking and unchecking weapons to enable in runs
* The UI also supports saving presets to use later
  * At this time, presets are automatically named using alphabetical letters
  * Once [feature/ui-textfield-widget](https://github.com/kartoFlane/ITB-ModLoader/tree/feature/ui-textfield-widget) is finished, it can be used to type in names for the presets. The UI is coded so that will be easy to accomplish, you can set the preset name in the config and it will show in the UI. Just change the logic [here](https://github.com/kartoFlane/ITB-ModLoader/commit/e88259ed743563601287cbf05e9a8f316a348cd8#diff-5c212bd8558c32cab90fc7b57a8dff6fR139-R155) to pull from a text field or a popup to get the name.

----

To make this work, I made two internal changes to the mod API:
* `ModsFirstLoadedHook`: fired after the first time mods are loaded. Fires after `ModsLoadedHook`, but will not fire when `load` is called later in the game, such as on starting runs.
* `ModsFirstLoadedHook` and `ModsInitializedHook` will no longer be cleared when other hooks are reset, allowing the mod loader to add logic to those times.


scripts/mod_loader/modui/weapon_deck_selector.lua
```diff
@@ -0,0 +1,470 @@
...
+		function disableAllButton.onclicked()
+			for _, button in ipairs(buttons) do
+				button.checked = false
+			end
+			enableSaveLoad(true)
+			return true
+		end
+
+		-- add spacer before preset buttons
+		-- width is crafted to right align the preset buttons
```
kartoFlane, line 301: Instead of creating a dummy spacer element, you should be able to simply set `alignH = "right"` on the button elements (`UiBoxLayout` feature)

scripts/mod_loader/modui/weapon_deck_selector.lua
```diff
@@ -0,0 +1,470 @@
...
+			table.sort(data.weapons, sortName)
+			table.insert(classList, data)
+		end
+		table.sort(classList, sortName)
+		-- create a frame for each class
+		local offset = 0
+		for _, class in ipairs(classList) do
+			-- 2 of the paddings is for the height, plus a little extra pading
+			local height = math.ceil(#class.weapons / weaponsPerRow) * CELL_HEIGHT + 4 * PADDING
+			local classArea = Ui()
```
kartoFlane, line 423: `UiFlowLayout` could probably simplify the layout here

scripts/mod_loader/modui/weapon_deck_selector.lua
```diff
@@ -0,0 +1,470 @@
...
+				loadPresetButton.disabled = value == PRESET_NEW
+			else
+				-- random can always load
+				savePresetButton.disabled = true
+				loadPresetButton.disabled = value ~= PRESET_RANDOM
+			end
+		end
+
+		-------------
+		-- Weapons --
```
kartoFlane, line 379: Generally it might be a good idea to separate such sections into functions - helps with readability

KnightMiner:
Finished addressing comments

PR #41, opened 2020-05-14
Improve pilot arrange UI and SDL2 update

KnightMiner:
This pull request makes some improvements to the pilot arrange UI to make it more like vanilla UIs. Specific changes:

* Pilots now show with a black silhouette and a lock when not yet unlocked, like in the vanilla UI
  * They will show no text on hover
  * They can still be reordered with other pilots
* Unlocked pilots now show their skill description in the tooltip
  * If the pilot has no skill, it shows "No Special Ability"
* If the hangar has been loaded, leaving the UI reminds the user to restart their game
  * I borrowed a few texts from the SDL restart reminder

In order to make this change not a pain to the user, first #40 should be merged so pilots unlock regardless of if they are included in the first 13 pilots.

----

To make this work, I needed to update SDL2. It now includes two new functions:
* `sdl.multiply(surface, color)`: Multiplies all pixels in a surface by the given color. Useful for:
  * Color masks. For example, in the pilot UI I mask the pilots using `sdl.rgb(0,0,0)` to blacken the portrait.
  * Tinting: can take a grayscale image and recolor it dynamically
  * Changing opacity: For example, multiplying by `sdl.rgba(255,255,255,128)` will reduce the opacity to 50%.
* `sdl.grayscale(surface)`: Converts an image to grayscale using the luminosity method.


scripts/mod_loader/modui/pilot_arrange.lua
```diff
@@ -65,6 +99,16 @@ local function createUi()
...
 		if updateImmediately then
 			PilotList = pilots
+		else
+			-- alert the user to restart the game
+			modApi:scheduleHook(50, function()
+				sdlext.showButtonDialog(
+					GetText("OpenGL_FrameTitle"),
+					GetText("PilotArrange_RestartWarning_Text"),
+					responseFn, nil, nil,
+					{ GetText("OpenGL_Button_Quit"), GetText("OpenGL_Button_Stay") }
```
kartoFlane, line 109: You probably shouldn't re-use the OpenGL dialog functionality here, as it's pretty specific in that it's a one-time thing that also prevents the whole mod loader from working unless the game is restarted.

Generally, other popups should follow the principle of being only informational, while also giving the option to prevent the popup from showing (eg. `modui/popup_script-error.lua`)

kartoFlane:
Can you also put up a PR with the changes to the [SDL repo](https://github.com/kartoFlane/IntoTheBreachLua)?

KnightMiner:
Currently I have everything pushed to my [fork](https://github.com/KnightMiner/IntoTheBreachLua) of the SDL repo. Did you want to continue maintaining that or do you want me to take it over? I wanted to get a proper readme explaining how to build the project so others don't spend as many hours dealing with it as me.

I pushed a commit styling the option like the mod restart warning. Currently, it shares that config, but if you want I can instead add a new modloader config for the pilot restart warning.

kartoFlane:
> I wanted to get a proper readme explaining how to build the project so others don't spend as many hours dealing with it as me.

A readme for that would be indeed much appreciated, I myself have put off digging into the SDL2 code in part due to that.

PR #43, opened 2020-11-01
Added code to overridden SkillEffect functions to account for new sig…

Lemonymous:
…natures

ITB 1.2 added new signatures for `AddProjectile` and `AddArtillery`.
Added code to account for these new signatures in overridden functions of the same names.
Also added compatibility code for weapons that used `piOrigin` before ITB 1.2

Note that `delay = PROJ_DELAY` no longer does anything. Feel free to suggest alterations on how to include it again.

### DIFF
--- a/scripts/mod_loader/modapi/skills.lua
+++ b/scripts/mod_loader/modapi/skills.lua
@@ -106,13 +106,26 @@ SkillEffect.AddGrapple = function(self, source, target, anim, ...)
     oldAddGrapple(self, source, target, anim, ...)
 end
 
+local function isUserdataPoint(var)
+	return type(var) == 'userdata' and type(var.x) == 'number' and type(var.y) == 'number'
+end
+
 local function overrideProjectileOrArtillery(funcName, oldFunc)
     local damageList = modApi:stringStartsWith(funcName, "AddQueued") and "q_effect" or "effect"
     local metadataType = funcName:gsub("^Add", ""):gsub("^Queued", ""):lower()
 
     assert(metadataType == "projectile" or metadataType == "artillery", "This function only works for projectile or artillery weapons")
 
-    SkillEffect[funcName] = function(self, damageInstance, projectileArt, delay, ...)
+    SkillEffect[funcName] = function(self, ...)
+		local origin, damageInstance, projectileArt, delay = ...
+		
+		if not isUserdataPoint(origin) then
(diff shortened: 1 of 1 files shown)


kartoFlane:
Neat, I completely forgot we had these overrides.

Does calling the new `AddProjectile`/`AddArtillery` override update the effect's `piOrigin` with the origin passed as first argument? If not, we might want to add that to the metadata table (defaulting to `piOrigin` for old effects).

For `piOrigin` set after the function is called, it might be possible to account for it with some combination of `skillBuildHook` and `Board:AddEffect` overrides, though it might be more trouble than it's worth. Not sure how common it is to set `piOrigin` after the call in old mods. Metadata and `__newindex` trickery probably won't work, since SkillEffect is a userdata object, not a table...

I was not aware of the `PROJ_DELAY` change. The intention of that line is to replicate default behaviour of projectiles, when the `delay` argument is omitted, so whatever achieves that goal, would be appropriate here.

Lemonymous:
`Does calling the new AddProjectile/AddArtillery override update the effect's piOrigin with the origin passed as first argument? If not, we might want to add that to the metadata table (defaulting to piOrigin for old effects).`
It does affect piOrigin, so yes, that would be a good idea.

`For piOrigin set after the function is called, ...`
It could work, but I am not sure it is worth the trouble.

`I was not aware of the PROJ_DELAY change. The intention of that line is to replicate default behaviour of projectiles, when the delay argument is omitted, so whatever achieves that goal, would be appropriate here.`
I assumed as much. I probably worded myself badly. What I meant was that _with my updated code_, that line will no longer do anything, as I never pass the variable `delay` into the following function and instead use `...`.
Both AddProjectile and AddArtillery have signatures to handle the case where delay is omitted though, so I am not sure the line is needed after all. Unless there is something else I am not aware of.

Lemonymous:
Actually, come to think of it, the 3 valid signatures are:
`void AddProjectile(SkillEffect&,Point,SpaceDamage,std::string,custom [float])`
`void AddProjectile(SkillEffect&,SpaceDamage,std::string)`
`void AddProjectile(SkillEffect&,SpaceDamage,std::string,custom [float])`
If signature 2 is called, and we try to convert it to signature 1; we need to add a default delay in every case, since there is no signature that accepts an origin without also specifying the delay.

PR #45, opened 2020-11-03
Added sorting to mod configuration

Lemonymous:
Added a dropdown button to the mod configuration window, able to sort mods by name, id or load order.

Sorting by name and id will sort every mod, enabled or not, according to sorting option.
Sorting by load order will sort every already enabled mod according to the order they were sorted at current mod loader initialization, while the remaining mods are sorted by id.
Mods enabled or disabled after mod loader init will not be sorted by load order, since they have not been loaded.

### DIFF
--- a/scripts/mod_loader/modui/mod_configuration.lua
+++ b/scripts/mod_loader/modui/mod_configuration.lua
@@ -140,6 +140,25 @@ local function createUi()
 			end
 			return false
 		end
+		
+		local line = Ui()
+				:width(1):heightpx(frametop.decorations[1].bordersize)
+				:decorate({ DecoSolid(frametop.decorations[1].bordercolor) })
+				:addTo(frametop)
+
+		local buttonHeight = 40
+		local buttonLayout = UiBoxLayout()
+				:hgap(20)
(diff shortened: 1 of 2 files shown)


scripts/mod_loader/modui/mod_configuration.lua
```diff
@@ -149,6 +168,50 @@ local function createUi()
...
+		
+		--- adds buttons to control sorting
+		local function addSortButtons()
+			-- dropdown to choose a sorting
+			local buttonLayoutRight = UiBoxLayout()
+				:hgap(20)
+				:heightpx(40)
+				:addTo(buttonLayout)
+			buttonLayoutRight.alignH = "right"
+			local presetDropdown = UiDropDown(sortDropDownOptions)
```
kartoFlane, line 185: Might  want to use `UiDropDown(values, strings, defaultValue)` constructor, so that you can compare selection to index, rather than exact strings.

Lemonymous: Makes sense; in case someone were to localize it.

PR #47, opened 2020-11-16
Misc

Lemonymous:
Miscellaneous changes and additions
Edit: Based on `assert` branch in different pull request

### DIFF
--- a/scripts/mod_loader/altered/misc.lua
+++ b/scripts/mod_loader/altered/misc.lua
@@ -7,8 +7,7 @@ function save_table(target, depth)
 		if type(v) == "table" then
 			if v ~= target.__index then
 				if not depth or depth > 0 then
-					depth = depth and depth - 1 or nil
-					value = save_table(v, depth)
+					value = save_table(v, depth and depth - 1 or nil)
 				else
 					value = "table"
 				end
--- a/scripts/mod_loader/altered/misc.lua
+++ b/scripts/mod_loader/altered/misc.lua
@@ -1,6 +1,48 @@
 
+-- modified vanilla save_table with optional 'depth' parameter
+function save_table(target, depth)
+	local ret = "{"
+	for i, v in pairs(target) do
+		local value = ""
+	    if type(v) == "table" then
+	        if v ~= target.__index then
+				if not depth or depth > 0 then
+					depth = depth and depth - 1 or nil
+					value = save_table(v, depth)
+				else
+					value = "table"
+				end
+			end
+		elseif type(v) == "userdata" and v["GetLuaString"] ~= nil then
+			value = v:GetLuaString()
+        elseif type(v) ~= "userdata" and type(v) ~= "function" then
+            if type(v) == "string" then
+	    	    value = "\""..v.."\""
+	    	elseif type(v) == "boolean" then
+	    	    value = v and "true" or "false"
+	        else
+                value = v
+            end
+	    end
+		
+		if type(i) == "string" then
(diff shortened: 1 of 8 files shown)


scripts/mod_loader/altered/misc.lua
```diff
@@ -1,6 +1,48 @@
...
+				if not depth or depth > 0 then
+					depth = depth and depth - 1 or nil
+					value = save_table(v, depth)
+				else
+					value = "table"
+				end
+			end
+		elseif type(v) == "userdata" and v["GetLuaString"] ~= nil then
+			value = v:GetLuaString()
+        elseif type(v) ~= "userdata" and type(v) ~= "function" then
```
KnightMiner, line 18: Seems to have some mixed tabs and spaces here. Github has that annoying "feature" where tabs show as 6 spaces, so it stands out. I think the rest of the file does tabs.

Lemonymous: Right you are! Maybe this "feature" is there to make it apparent.
It also revealed a bug when testing the function again, so that's an extra bonus

kartoFlane:
After merging #46, you now have conflicts - there are two solutions here:
1. Merge the `master` into `misc` and resolve conflicts.
This is the simplest and most common solution, but the commit graph looks ugly (also potentially leads to some other nastiness with git I don't understand enough to explain well)
2. Rebase your entire `misc` branch back onto mod loader's `master` branch, and force-push changes here.
The benefit of this approach is particularly neat in the case of this PR, where you're based off of a branch that's already been merged. Rebasing will "trim" the commits that are part of this PR, but are already present on the target branch.
(generally, this should be done before you push your branch online and submit a PR, as you're literally rewriting git history, which is really bad if someone else is also working on your branch, 'cause then *they* also have to rebase their own changes & potentially resolve further conflicts (or delete the branch and start over) - but if you're the sole developer, then it's fine for the most part)

Lemonymous:
Aha, no problem! I suspected there might be issues with merging them, but I opened the pull request anyways, just to see how git would handle it. Hopefully it was harmless enough.
I'll get it sorted.

scripts/mod_loader/modapi/map.lua
```diff
@@ -15,6 +15,19 @@ function modApi:fileExists(name)
 	end
 end
 
+function modApi:directoryExists(path)
+	if type(path) ~= "string" then return false end
+
+	local response = os.execute("cd " .. path)
```
kartoFlane, line 21: I know I used it myself, but... `os.execute` causes the command line terminal to appear briefly, which is not only ugly and startling for non-tech savvy players, but also somewhat slow, since the operating system has to start up a new terminal to execute the command. 
Generally I'd advise against using it in API functions - rather, limit its usage to internal mod loader and game startup only, if it's absolutely required.

Luckily, it's possible to use another function from `os` here: `os.rename` ([see example](https://stackoverflow.com/a/40195356)), which doesn't cause the terminal to pop up and is *much* faster.

`modApi:fileExists` will also benefit from this, since we won't have to try to open the file for reading.

Lemonymous: Ah, I was not aware. That would indeed be ugly.

With that change; `fileExists` will return true for both files and folders, while `directoryExists` will return true only for folders.
Would you consider that correct behavior?

kartoFlane: Yes, that'd be correct, given the tools we have to work with.

PR #50, opened 2020-12-15
Miscellaneous additions and changes

Lemonymous:
A list of additional functionality for future features, which should leave everything current as it is.

### DIFF
--- a/scripts/mod_loader/modapi/ftldat.lua
+++ b/scripts/mod_loader/modapi/ftldat.lua
@@ -3,6 +3,19 @@
 
 local FtlDat = require("scripts/mod_loader/ftldat/ftldat")
 
+function modApi:assetExists(resource)
+	Assert.ResourceDatIsOpen("writeAsset")
+	assert(type(resource) == "string")
+	
+	for i, file in ipairs(self.resource._files) do
+		if file._meta._filename == resource then
+			return true
+		end
+	end
+	
+	return false
+end
+
 --[[
 	Writes a new file to the resource.dat archive with the specified content.
 
--- a/scripts/mod_loader/modapi/ftldat.lua
+++ b/scripts/mod_loader/modapi/ftldat.lua
@@ -4,7 +4,7 @@
 local FtlDat = require("scripts/mod_loader/ftldat/ftldat")
 
 function modApi:assetExists(resource)
-	Assert.ResourceDatIsOpen("writeAsset")
+	Assert.ResourceDatIsOpen("assetExists")
 	assert(type(resource) == "string")
 	
 	for i, file in ipairs(self.resource._files) do
--- a/scripts/mod_loader/modapi/vector.lua
+++ b/scripts/mod_loader/modapi/vector.lua
@@ -1,0 +1,306 @@
+
+math.vec2 = {}
+math.vec3 = {}
+math.vec4 = {}
+
+local mt_vec2 = {
+	x = 1, y = 2,
+	[1] = 1, [2] = 2
+}
+
+local mt_vec3 = {
+	x = 1, y = 2, z = 3,
+	i = 1, j = 2, k = 3,
+	[1] = 1, [2] = 2, [3] = 3
+}
+
+local mt_vec4 = {
+	top = 1, bottom = 2, left = 3, right = 4,
+	north = 1, south = 2, west = 3, east = 4,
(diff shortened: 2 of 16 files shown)


scripts/mod_loader/modapi/ftldat.lua
```diff
@@ -3,6 +3,19 @@
 
 local FtlDat = require("scripts/mod_loader/ftldat/ftldat")
 
+function modApi:assetExists(resource)
+	Assert.ResourceDatIsOpen("writeAsset")
```
kartoFlane, line 7:
```suggestion
	Assert.ResourceDatIsOpen("assetExists")
```

scripts/mod_loader/modapi/vector.lua
```diff
@@ -0,0 +1,380 @@
...
+		)
+		
+	elseif #a == 2 or #b == 2 then
+		return math.vec2(
+			a[1] + b[1],
+			a[2] + b[2]
+		)
+	end
+	
+	return nil
```
kartoFlane, line 76: Maybe throw an error? It might be difficult for people to debug that they're trying to add two vectors of different sizes.

KnightMiner: If I am reading the code right, adding vectors of different sizes will return the larger size. This nil case means you had vectors of sizes less than 2 or greater than 4.

scripts/mod_loader/tests/vector.lua
```diff
@@ -0,0 +1,229 @@
...
+
+function testsuite.test_vector2_xy()
+	local v = math.vec2(11,22)
+
+	Assert.TableEquals({11,22}, v, "math.vec2(x, y)")
+	
+	return true
+end
+
+function testsuite.test_vector2_v2()
```
kartoFlane, line 11: Some of the test names here aren't very descriptive; you might want to consider changing some of them:

current | suggested
------------ | -------------
`test_vector2_v2` | `test_vector2_copy_constructor`
`test_vector3_v2z` | `test_vector3_constructor_from_vector2`

...and so on.

These functions aren't manually called by anyone, so it's okay if they're somewhat long-winded (unless you invoke them by hand from the in-game lua console; personally I quickly got tired of that and created the testing window :) ).

Ideally, test names should be clear enough for someone without any knowledge of their implementation to know what they're testing (or know what they're supposed to test, so that they can be fixed in the event that they're broken) - this is not always easy :)

scripts/mod_loader/modapi/vector.lua
```diff
@@ -0,0 +1,306 @@
...
+end
+
+local function subVectors(a, b)
+	return addVectors(a, b(-b[1], -b[2], -b[3], -b[4]))
```
KnightMiner, line 80: I feel using `-b` would be better in this case, overriding `__unm` in the metatable so `-b` negates the vector.

scripts/mod_loader/modapi/vector.lua
kartoFlane: Ah, you're right - I blindly assumed those would be `and` guards, not `or` guards

scripts/mod_loader/modapi/vector.lua
Lemonymous: I totally agree - I Didn't know __unm existed.

scripts/mod_loader/tests/vector.lua
Lemonymous: That makes sense. I guess I got lazy when naming them to start testing and never revisited.

scripts/mod_loader/modapi/vector.lua
Lemonymous: I guess throwing error wouldn't hurt either way.

scripts/mod_loader/modapi/vector.lua
Lemonymous: Added Unary Minus

scripts/mod_loader/tests/vector.lua
Lemonymous: Renamed tests to be more descriptive

scripts/mod_loader/modapi/vector.lua
Lemonymous: Changed `return nil` to throw an error instead

kartoFlane (approved):

PR #52, opened 2021-01-21
Fixes/tests

Lemonymous:
Test runner crashed the game when a test failed. This should alleviate the issue.


kartoFlane:
The irony of all tests passing, leading to an undetected bug in test running logic... 😄

PR #57, opened 2021-02-07
Feature modpack support

Lemonymous:
Mods can now set an optional field `submodFolders`, which is a `string` or `table of strings` of paths to folders with nested mods in them.

`Mod configuration` has been changed to display nested mods in a tree-like structure.
Additional assets and helper functions has been added to support this feature.

Current issues:
- [Solved in #60] flickering when opening and closing nested mods.
- height of the mod content frame is set to contain the height of the main mods, and can be too narrow if you have just a few modpacks with many nested mods in them.
- sorting by id and name does not appear to actually sort the mods correctly. (This bug is also present on the master branch)


kartoFlane:
Remaining issues will be handled in subsequent PRs

PR #59, opened 2021-02-08
Adding the possibility to define pawn type's moveSkill as a string

Lemonymous:
Most (if not all) of a pawn type's definition is made of numbers, booleans and strings.

This change allows `moveSkill` to also be defined as a string, instead of requiring a direct reference to the skill table itself.


kartoFlane:
It might make sense to also have that code check whether a table/skill with that name exists, and throw an error if it doesn't. As it stands now it will silently use the default Move skill, which might be confusing to debug.

PR #61, opened 2021-02-11
Implement palette support + UI

Lemonymous:
Adds native palette support to the mod loader, as well as ui to rearrange usable palettes in the mech hangar.
Adding palettes with the new mod loader functions, FURL, KnightMiner's custom palette library or Lemonymous' palette library will all have their palettes automatically added to the ui.

No additional action is required by modders.

Adds two new main functions to use by mods:
`void modApi:addPalette(palette)`
`int modApi:getPaletteImageOffset(id)`

Adds functions to be used by the palette ui:
`table modApi:getCurrentPaletteOrder()`
`table modApi:getDefaultPaletteOrder()`
`table modApi:getPalette(id)`

Adds additional helper functions to modApi:
`table modApi.getColorMap(paletteIndex)` - same as `GetColorMap`
`int modApi.getColorCount()` - same as `GetColorCount`
`bool modApi:canAddPalettes()`

Known issues:
- Draggable ui elements in the palette ui does not respond correctly to clicking the right mouse button outside of the dragged element. This is a bigger issue with the mod loader's draggable code and is probably beyond the scope of this PR.


scripts/mod_loader/modapi/compat.lua
```diff
@@ -64,3 +64,18 @@ local statusMetatable = {
 	end
 }
 STATUS_TOOLTIPS = setmetatable(STATUS_TOOLTIPS, statusMetatable)
+
+-- ///////////////////////////////////////////////////////////////////////////////////
+-- compatibility code for KnightMiner's customPalettes
+local palettes = { VERSION = "1" }
+CUSTOM_PALETTES = palettes
```
KnightMiner, line 71: This is a little incomplete I think, as it won't prevent my library from calling it's finalize hook. I think there is just an internal boolean you need to set. I'll have to look at this more later to see if anything else is needed.

scripts/mod_loader/modapi/palette.lua
```diff
@@ -0,0 +1,354 @@
...
+	end
+	
+	for i = 1, 8 do
+		Assert.Equals('userdata', type(new_palette[i]), string.format("Argument #1 - incomplete palette"))
+	end
+	
+	palettes:addPalette(id, new_palette)
+end
+
+function modApi:getPaletteIndex(id)
```
KnightMiner, line 331: I'm not sure I like calling this `getPaletteIndex`, since that typically means the true index. This function is closer to `getPaletteImageOffset`, makes it clear it uses a different indexing scheme.

scripts/mod_loader/modapi/palette.lua
```diff
@@ -0,0 +1,354 @@
...
+	
+	getPalette = function(self, index)
+		local hash = self.hash[index]
+		if not hash then return nil end
+		
+		return self.palette[hash]
+	end,
+	
+	getCount = function(self)
+		return #self.hash
```
KnightMiner, line 28: My version of this function looked like:
```lua
    local count = #self.hash
    if count > 11 and sdlext.isHangar() then
      return 11
    end
    return count
```

Effectively it caused only 11 palettes to show in the hangar instead of showing extra ones with broken buttons, similar to the pilot patch

Lemonymous: I'm not sure I understand what you mean by "extra ones with broken buttons"

scripts/mod_loader/modapi/palette.lua
Lemonymous: I think I can get behind that.

scripts/mod_loader/modapi/palette.lua
```diff
@@ -0,0 +1,433 @@
...
+		if type(a) == 'number' then
+			return self._array[a]
+		else
+			return self._dictionary[a]
+		end
+	end,
+
+	swap = function(self, a, b)
+		local paletteA, paletteB
+		if type(a) == 'number' then
```
kartoFlane, line 74: Shouldn't you use the `get` function defined above here? No reason to duplicate the logic

scripts/mod_loader/modapi/palette.lua
```diff
@@ -0,0 +1,433 @@
...
+		Assert.Equals('userdata', type(gl_colorMap[i]), string.format("Argument #1 - incomplete colorMap"))
+	end
+
+	return gl_colorMap
+end
+
+local function buildPaletteImages(palette)
+	local images = palette.images or palette.Images
+
+	Assert.Equals({'nil', 'table'}, type(images), string.format("images of palette", i))
```
kartoFlane, line 223: Looks like `i` is undefined here

scripts/mod_loader/modui/palette_arrange.lua
```diff
@@ -0,0 +1,355 @@
...
+
+		for i = 2, #content.children do
+			local ui = content.children[i]
+			if rect_contains(ui.rect, mx, my) then
+				if ui ~= placeholder then
+					table.remove(content.children, list_indexof(content.children, placeholder))
+					table.insert(content.children, i, placeholder)
+				end
+
+				if #content.children > 11 then
```
kartoFlane, line 135: Since the constant `11` appears several times throughout, it might make sense to put it in a variable, similar to how pilot arrange screen defines `MAX_PILOTS`

Lemonymous: I went with file-locals. The constants will likely not have to change unless ITB gets an expansion, in which case everything will have to be reviewed anyways.

PR #64, opened 2021-02-14
Update dialog helper functions, rebuild mod content screens

kartoFlane:
- Change base `Ui:relayout()` to not reset percentage size fields - should make the function more useful for updating UI in some contexts.
- Changed signature of dialog helper functions to accept an `options` argument that allows greater control over their behaviour
  - Updated mod configuration dialog to utilize this to fix an issue with the dialog being extremely tiny when opened with very few mods (eg. one)
  - Attempts to use old signatures in mods should still work
- Rebuilt dialog helper functions using layouts
- Rebuilt pilot arrange screen using layouts
- Rebuilt mod content screen using dialog helpers
- Rebuilt weapon deck selector


scripts/mod_loader/modui/dialog_helper.lua
```diff
@@ -136,25 +136,36 @@ end
...
+		:orientation(false)
+		:decorate({ DecoFrameHeader(), DecoFrame() })
+		:caption(title)
 
 	frame.scroll = UiScrollArea()
-			:width(1):height(1)
-			:padding(10)
-			:addTo(frame)
+		:width(1):height(1)
+		:padding(10)
```
Lemonymous, line 158: Maybe it would be an idea to have an option to define padding when creating the dialog?
Padding cannot realistically be changed later, since other elements are based on this value.

PR #68, opened 2021-02-22
Add events for when opening/closing various windows

Lemonymous:
Issue: (at least) on my version of the game, the mod loader features in the hangar are not very accurate, and often fails.
Example: Opening/closing the Squad Selection window can often make the extra mod loader difficulties, appear when the Squad Selection Window is open; and disappear when it is closed. I believe the current way of detecting the screen is by calculating the size of the topmost window; which may be inaccurate with my settings of the game.

This PR is an attempt to provide a different method to detect which window is currently open. This method checks when the game requests the text of the title of the opened window. So in order for this to function without getting false positives, `GetText` should not be called for the titles of those windows unless called by the ITB game itself.

In the process of creating this, I also added events for most other windows that I found unique texts for, which may or may not be useful for other things.

Adds the following events in `modApi.event`:
`onEscapeMenuWindowShown`
`onEscapeMenuWindowHidden`
`onSquadSelectionWindowShown`
`onSquadSelectionWindowHidden`
`onPilotSelectionWindowShown`
`onPilotSelectionWindowHidden`
`onAchievementsWindowShown`
`onAchievementsWindowHidden`
`onOptionsWindowShown`
`onOptionsWindowHidden`
`onLanguageSelectionWindowShown`
`onLanguageSelectionWindowHidden`
`onHotkeyConfigurationWindowShown`
`onHotkeyConfigurationWindowHidden`
`onProfileSelectionWindowShown`
`onProfileSelectionWindowHidden`
`onCreateProfileConfirmationWindowShown`
`onCreateProfileConfirmationWindowHidden`
`onDeleteProfileConfirmationWindowShown`
`onDeleteProfileConfirmationWindowHidden`
`onStatisticsWindowShown`
`onStatisticsWindowHidden`
`onNewGameWindowShown`
`onNewGameWindowHidden`
`onAbandonTimelineWindowShown`
`onAbandonTimelineWindowHidden`

Also adds the following functions:
`sdlext.isEscapeMenuWindowVisible`
`sdlext.isSquadSelectionWindowVisible`
`sdlext.isAchievementsWindowVisible`
`sdlext.isPilotSelectionWindowVisible`
`sdlext.isOptionsWindowVisible`
`sdlext.isLanguageSelectionWindowVisible`
`sdlext.isHotkeyConfigurationWindowVisible`
`sdlext.isProfileSelectionWindowVisible`
`sdlext.isCreateProfileConfirmationWindowVisible`
`sdlext.isDeleteProfileConfirmationWindowVisible`
`sdlext.isStatisticsWindowVisible`
`sdlext.isNewGameWindowVisible`
`sdlext.isAbandonTimelineWindowVisible`


kartoFlane:
I like the idea overall, but I'd prefer we move away from referencing mod-loader defined fields via constructed strings, ie.:
```lua
modApi.events[string.format("on%sEntered", self.event_id)]:dispatch()
```
This technique *is* fairly convenient and helpful for building reusable code, but unfortunately it makes it nearly impossible to find usages of referenced fields, either via global search or tools like "find references" in IDEs.

While it might not be much of an issue for other use cases, I think it's particularly important for events, since they will tie many disassociated systems together, so traceability is very useful to understanding how and where the event is used.

To account for this without bloating the code, I'd suggest changing the windows definitions as follows:
```lua
function Window:open()
	self.opened = true
	table.insert(openWindows, self)
	self.event_open:dispatch()
end

function Window:close()
	self.opened = false
	remove_element(self, openWindows)
	self.event_close:dispatch()
end

-- ...

Escape_Title = Window:new{
	event_open = modApi.events.onEscapeMenuEntered, -- direct reference to the field that allows for traceability
	event_close = modApi.events.onEscapeMenuExited,
},
-- etc...
```

PR #76, opened 2021-04-13
Made scrolling of UiScrollArea smoother

Lemonymous:
Added variables:
- dyTarget: dy (which is the scroll offset of the ui element), will attempt to close in towards dyTarget each frame.
- scrollSpeed: The number of pixels each scroll click will yield.
- scrollOvershoot: The number of pixels we can overshoot the top/bottom of the ui element. It will always bounce back to the content of the ui element once scrolling stops.
- scrollChangefactor: The rate at which dy approaches dyTarget by the formula: dy = dy * (1-factor) + dyTarget * factor. So factor 0 will never change anything, while factor 1 will instantly change dy to dyTarget.

### DIFF
--- a/scripts/mod_loader/ui/widgets/scrollarea.lua
+++ b/scripts/mod_loader/ui/widgets/scrollarea.lua
@@ -12,6 +12,10 @@ function UiScrollArea:new()
 	
 	self.padr = self.padr + self.scrollwidth
 	self.nofity = true
+	self.dyTarget = 0
+	self.scrollSpeed = 100
+	self.scrollOvershoot = 0
+	self.scrollChangefactor = 0.4 -- <0.0, 1.0]
 
 	self.scrollPressed = false
 	self.scrollHovered = false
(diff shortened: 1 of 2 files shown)


scripts/mod_loader/ui/widgets/scrollarea.lua
```diff
@@ -105,35 +121,9 @@ function UiScrollArea:mouseup(x, y, button)
 	return Ui.mouseup(self, x, y, button)
 end
 
-function UiScrollArea:computeOffset(scrollAmount)
```
kartoFlane, line 108: That function is also used by `pilot_arrange.lua:184` to compute correction offset when scrolling the window while dragging a pilot. I think it's better to keep this function around, since them we can find the offset without actually scrolling the element. If left as-is, the pilot window will crash when the user attempts to scroll.

kartoFlane (changes requested):

scripts/mod_loader/ui/widgets/scrollarea.lua
Lemonymous: I see. I was not aware of this.
In my test when I set `scrollOvershoot = 100`, I was able to use the mousewheel to scroll all the current mod content ui entries, including the arrange pilot ui, without crashing. I suppose it might come into effect if I had so many additional modded pilots that the window was too little to contain them and then using the scrollbar on the right side?

Lemonymous: I created 100 pilots in a mod to enable the scrollbar, but I am still not getting any crash when using either the mouse wheel or dragging the scrollbar. The function `button.dragWheel` on line 183 in file `pilot_arrange.lua` does not appear to run in my tests.

Lemonymous: I'm actually not sure what the right solution would be. The way I changed UiScrollArea.wheel, makes it so that function itself does not scroll any amount It only prompts UiScrollArea.relayout to make the adjustments to `dy` on the following frames. As far as I understand it, making any adjustments to the droptargets in PilotArrange on dragWheel would be premature.

However, it appears the following code in UiScrollArea.relayout is not sufficient to make the proper adjustments to droptargets in PilotArrange.
```lua
if doUpdate then
    Ui.mousemove(self, sdl.mouse.x(), sdl.mouse.y())
end
```

Lemonymous: The following hack would update the elements in PilotArrange. I like it even less than the almost as bad hack as calling Ui.mousemove outside of mousemove functions though.
```lua
if doUpdate then
	self.root:event{ type = function(self) return sdl.events.mousemotion end }
end
```

kartoFlane: Alright, the fix turned out to be fairly simple. Still slightly hacky, but not nearly as much :P

kartoFlane (approved):

PR #78, opened 2021-04-17
Rework UI system

Lemonymous:
This is an attempt to streamline the ui system.

## Dropdowns
Are now added to a separate dropdownUi element in root, and has automatic cleanup on mousedown events if they ever get orphaned; but otherwise don't need any more special treatment.

## Draggable
Can now be added to a separate draggableUi element in root, in order for it to be drawn on top of most everything else (barring dropdowns and tooltips). Some functionality has also been added to UiDraggable, in order to streamline the process of creating windows that let you drag an element among a list of elements.

## UiRoot:updateStates()
This function runs each frame in order to update important variables in every descendant.

&nbsp;
## Ui.containsMouse
This variable is `true` if the element contains the mouse cursor; otherwise `false`
- This is ensured by the function `Ui.updateContainsMouse`

## Ui:updateContainsMouse(mx, my)
This function goes through every descendant ui element in order to update their `containsMouse` state.
- `containsMouse` is set to `true` if; the element contains the mouse and is `visible` and `not ignoreMouse`; otherwise false.
- `mouseEntered` or `mouseEntered` is called for the element when `containsMouse` state changes.
- When called from root, it is ensured to go through every element to make sure `containsMouse` is always up to date, as long as no ui element has overridden `updateContainsMouse` to break the search.

&nbsp;

## Ui.hovered
This variable should only ever be true for one element. It is the topmost hoverable element contained by the mouse cursor, _or_ the pressed element.
- In order for an element to be hoverable it must be `visible`, `not ignoreMouse`, `containsMouse` and `not translucent`
- However, a `translucent` element may be the ancestor of a hoverable element.

## Ui:updateHoveredState()
This function goes through every descendant ui element in order to update their `hovered` state.
- `UiRoot.updateHoveredState` is a separate function that resets the `hovered` state, and initiates the search for the hovered element by calling `Ui.updateHoveredState(root)`
- When iterating children, we know the first hoverable element we find will either be the first drawn element, or the ancestor of the first drawn element. With this knowledge, we can exit early, which means not all elements will be iterated.

&nbsp;

## UiRoot:updateDraggedState()
This function is kind of a hack. It checks if there is a `pressed` element, and if it is `draggable`. If there is, it calls `dragMove(mx, my)` for that element. It ensures dragged elements can make updates if the ui around them changes, even in the absence of mousemove events. There could possibly be another function called `dragUpdate` or similar to do the same task.
- There is no equivalent `Ui.updateDraggedState()` function

&nbsp;

## Ui:updateTooltipState()
This function is called for the `hovered` element, in order for that element to update the tooltip however it sees fit.
- `UiRoot.UpdateTooltipState` is a separate function that resets the tooltip, and calls `UpdateTooltipState` for the `hovered` element.

&nbsp;

## Ui:wheel()
If there is a `pressed` element, then `wheel` (and potentially `dragWheel`) is called for this element, and the mouse event is consumed if any of them returns `true`.

If the event was not consumed, then this function will be called for every ui element, as long as all of their ancestors are `visible`, `not ignoreMouse`, `containsMouse`. If any element returns `true`, the mouse event will be consumed, and the game will not execute this event in the vanilla game.
- `translucent` elements return `false`; otherwise `true`

&nbsp;

## Ui:mousedown() [**Edited**]
If the button is the left mousebutton, and there is a `hovered` element, `mousedown` will _only_ be called for this element (as well as any descendants), regardless of whether they consume the mouse event or not.

There is one exception to the first rule. Dropdowns that do not contain the mouse will recieve a `mousedown` call. This is to allow them to close themselves. Since this is an exception, their return value will not affect whether the mouse event is consumed or not.

If the event was not consumed, then this function will be called for every ui element, as long as all of their ancestors are `visible`, `not ignoreMouse`, `containsMouse`. If any element returns `true`, the mouse event will be consumed, and the game will not execute this event in the vanilla game.
- `translucent` elements return `false`; otherwise `true`

&nbsp;

## Ui:mouseup() [**Edited**]
If the button is the left mousebutton, and there is a `pressed` element, `mouseup` (and potentially `clicked`) will _only_ be called for this element (as well as any descendants), regardless of whether they consume the mouse event or not.

Otherwise, this function will be called for every ui element, as long as all of their ancestors are `visible`, `not ignoreMouse`, `containsMouse`. If any element returns `true`, the mouse event will be consumed, and the game will not execute this event in the vanilla game.
- `translucent` elements return `false`; otherwise `true`

&nbsp;

## Ui:mousemove()
If there is a `pressed` element, then `mousemove` (and potentially `dragMove`) is called for this element, and the mouse event is consumed if any of them returns `true`.

If the event was not consumed, then this function will be called for every ui element, as long as all of their ancestors are `visible`, `not ignoreMouse`, `containsMouse`. If any element returns `true`, the mouse event will be consumed, and the game will not execute this event in the vanilla game.
- `translucent` elements return `false`; otherwise `true`

### DIFF
--- a/scripts/mod_loader/ui/widgets/draggable.lua
+++ b/scripts/mod_loader/ui/widgets/draggable.lua
@@ -198,12 +198,14 @@ function UiDraggable:dragMove(mx, my)
 			self.h = math.max(self.dragH + my - self.dragY, minsize)
 		end
 	elseif self.dragMovable and self.dragMoving then
-		self.x = self.x + mx - self.dragX
-		self.y = self.y + my - self.dragY
+		local diffx = mx - self.dragX
+		local diffy = my - self.dragY
+		self.x = self.x + diffx
+		self.y = self.y + diffy
+		self.screenx = self.screenx + diffx
+		self.screeny = self.screeny + diffy
 		self.dragX = mx
 		self.dragY = my
-		self.screenx = self.x
-		self.screeny = self.y
 
 		self:processDropTargets(mx, my)
 	else
--- a/scripts/mod_loader/ui/widgets/draggable.lua
+++ b/scripts/mod_loader/ui/widgets/draggable.lua
@@ -91,10 +91,6 @@ function UiDraggable:startDrag(mx, my, button)
 		self.x = self.screenx
 		self.y = self.screeny
 	end
-
-	if self.parent then
-		self.parent:relayout()
-	end
 end
 
 function UiDraggable:stopDrag(mx, my, button)
--- a/scripts/mod_loader/ui/widgets/draggable.lua
+++ b/scripts/mod_loader/ui/widgets/draggable.lua
@@ -76,6 +76,22 @@ function UiDraggable:startDrag(mx, my, button)
 		self.__resizeDir = getResizeDirection(self, mx, my, self.__resizeHandle)
 	end
 
+	if self.dragPlaceholder ~= nil then
+		local root = sdlext.getUiRoot()
+		local owner = self.parent
+		local index = list_indexof(owner.children, self)
+		self:detach()
+		self.owner = owner
+		self.translucent = true
+
+		owner:add(self.dragPlaceholder, index)
+		self.dragPlaceholder:show()
+
+		self:addTo(root.draggableUi)
+		self.x = self.screenx
+		self.y = self.screeny
+	end
+
 	if self.parent then
 		self.parent:relayout()
 	end
@@ -86,6 +102,16 @@ function UiDraggable:stopDrag(mx, my, button)
 
 	if button ~= 1 then
 		return
+	end
+
+	if self.dragMoving and self.dragPlaceholder ~= nil then
+		local index = list_indexof(self.owner.children, self.dragPlaceholder)
+		self.dragPlaceholder:detach()
+		self.dragPlaceholder:hide()
+
+		self:detach()
+		self:addTo(self.owner, index)
+		self.translucent = self.dragPlaceholder.translucent
 	end
 
 	self.dragMoving   = false
@@ -110,14 +136,18 @@ function UiDraggable:stopDrag(mx, my, button)
 				self:onDraggableDropped(self, target)
 			end
 			
-			target.hovered = false
 			self.hoverTarget = nil
 		end
 	end
 end
 
+function UiDraggable:getDropTargets()
+	return self.dropTargets
+end
+
 function UiDraggable:processDropTargets(mx, my)
-	if not self.dropTargets then
+	local dropTargets = self:getDropTargets()
+	if not dropTargets then
 		return
 	end
 
@@ -127,7 +157,6 @@ function UiDraggable:processDropTargets(mx, my)
 	if target then
 		if not rect_contains(target.rect, mx, my) then
 			self.hoverTarget = nil
-			target.hovered = false
 
 			if target.onDraggableExited then
 				target:onDraggableExited(self, target)
@@ -136,11 +165,10 @@ function UiDraggable:processDropTargets(mx, my)
 		first = 2
 	end
 
-	for i = first, #self.dropTargets do
-		local target = self.dropTargets[i]
-		if rect_contains(target.rect, mx, my) then
+	for i = first, #dropTargets do
+		local target = dropTargets[i]
+		if self ~= target and rect_contains(target.rect, mx, my) then
 			self.hoverTarget = target
-			target.hovered = true
 
 			if target.onDraggableEntered then
 				target:onDraggableEntered(self, target)
@@ -151,7 +179,7 @@ function UiDraggable:processDropTargets(mx, my)
 			end
 
 			-- swap to front, to save iterations later
-			self.dropTargets[1], self.dropTargets[i] = self.dropTargets[i], self.dropTargets[1]
+			dropTargets[1], dropTargets[i] = dropTargets[i], dropTargets[1]
 
 			break
 		end
@@ -200,6 +228,8 @@ function UiDraggable:dragMove(mx, my)
 		self.y = self.y + my - self.dragY
 		self.dragX = mx
 		self.dragY = my
+		self.screenx = self.x
+		self.screeny = self.y
 
 		self:processDropTargets(mx, my)
 	else
@@ -213,6 +243,8 @@ function UiDraggable:dragWheel(mx, my, y)
 	else
 		self.__index.dragWheel(self, mx, my, y)
 	end
+
+	return false
 end
 
 function UiDraggable:mousemove(mx, my)
@@ -240,12 +272,17 @@ local function registerDragFunctions(self)
 	self.mousemove          = UiDraggable.mousemove
 	self.mouseExited        = UiDraggable.mouseExited
 	self.processDropTargets = UiDraggable.processDropTargets
+	self.getDropTargets     = UiDraggable.getDropTargets
 end
 
 function Ui:registerDragMove()
 	registerDragFunctions(self)
 	self.draggable   = true
 	self.dragMovable = true
+end
+
+function Ui:registerDragPlaceholder(placeholder)
+	self.dragPlaceholder = placeholder
 end
 
 function Ui:registerDragResize(resizeHandleSize, minSize)
@@ -257,7 +294,7 @@ function Ui:registerDragResize(resizeHandleSize, minSize)
 end
 
 function Ui:registerDropTarget(target)
-	self.dropTargets = self.dropTargets or {}
+	self.dropTargets = self:getDropTargets() or {}
 	table.insert(self.dropTargets, target)
 end
 
--- a/scripts/mod_loader/ui/widgets/draggable.lua
+++ b/scripts/mod_loader/ui/widgets/draggable.lua
@@ -75,22 +75,6 @@ function UiDraggable:startDrag(mx, my, button)
 		self.dragResizing = UiDraggable.isEdge(self, mx, my, self.__resizeHandle)
 		self.__resizeDir = getResizeDirection(self, mx, my, self.__resizeHandle)
 	end
-
-	if self.dragPlaceholder ~= nil then
-		local root = sdlext.getUiRoot()
-		local owner = self.parent
-		local index = list_indexof(owner.children, self)
-		self:detach()
-		self.owner = owner
-		self.translucent = true
-
-		owner:add(self.dragPlaceholder, index)
-		self.dragPlaceholder:show()
-
-		self:addTo(root.draggableUi)
-		self.x = self.screenx
-		self.y = self.screeny
-	end
 end
 
 function UiDraggable:stopDrag(mx, my, button)
@@ -98,16 +82,6 @@ function UiDraggable:stopDrag(mx, my, button)
 
 	if button ~= 1 then
 		return
-	end
-
-	if self.dragMoving and self.dragPlaceholder ~= nil then
-		local index = list_indexof(self.owner.children, self.dragPlaceholder)
-		self.dragPlaceholder:detach()
-		self.dragPlaceholder:hide()
-
-		self:detach()
-		self:addTo(self.owner, index)
-		self.translucent = self.dragPlaceholder.translucent
 	end
 
 	self.dragMoving   = false
@@ -275,10 +249,6 @@ function Ui:registerDragMove()
 	registerDragFunctions(self)
 	self.draggable   = true
 	self.dragMovable = true
-end
-
-function Ui:registerDragPlaceholder(placeholder)
-	self.dragPlaceholder = placeholder
 end
 
 function Ui:registerDragResize(resizeHandleSize, minSize)
--- a/scripts/mod_loader/ui/widgets/draggable.lua
+++ b/scripts/mod_loader/ui/widgets/draggable.lua
@@ -129,7 +129,7 @@ function UiDraggable:processDropTargets(mx, my)
 	local target = self.hoverTarget
 
 	if target then
-		if not rect_contains(target.rect, mx, my) then
+		if not target.containsMouse then
 			self.hoverTarget = nil
 
 			if target.onDraggableExited then
@@ -141,7 +141,7 @@ function UiDraggable:processDropTargets(mx, my)
 
 	for i = first, #dropTargets do
 		local target = dropTargets[i]
-		if self ~= target and rect_contains(target.rect, mx, my) then
+		if self ~= target and target.containsMouse then
 			self.hoverTarget = target
 
 			if target.onDraggableEntered then
--- a/scripts/mod_loader/ui/widgets/draggable.lua
+++ b/scripts/mod_loader/ui/widgets/draggable.lua
@@ -11,6 +11,10 @@ function UiDraggable:new()
 end
 
 function UiDraggable:isEdge(mx, my)
+	if not self.containsMouse then
+		return false
+	end
+
 	local resizeHandle = self.__resizeHandle
 	return mx < self.screenx + resizeHandle          or
 	       my < self.screeny + resizeHandle          or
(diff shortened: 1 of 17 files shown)


scripts/mod_loader/ui/widgets/root.lua
```diff
@@ -84,79 +196,76 @@ function UiRoot:event(eventloop)
...
+		for _, dropdown in ipairs(self.dropdownUi.children) do
+			if not dropdown.containsMouse then
+				dropdown:mousedown(mx, my, button)
+			end
 		end
-		return done
+
+		self:cleanupDropdown()
+
+		return consumeEvent
```
kartoFlane, line 239: So if I understand the above code right, `Ui:mousedown` will never be called for mouse buttons other than 1, and it won't be possible for UI to react to `mousedown` events for buttons 2, 3, 4 and 5?

scripts/mod_loader/ui/widgets/draggable.lua
```diff
@@ -76,8 +76,20 @@ function UiDraggable:startDrag(mx, my, button)
...
+		self:detach()
+		self.owner = owner
+		self.translucent = true
+
+		owner:add(self.dragPlaceholder, index)
+		self.dragPlaceholder:show()
+
+		self:addTo(root.draggableUi)
+		self.x = self.screenx
+		self.y = self.screeny
```
kartoFlane, line 92: I'm not entirely sold on the idea of placeholder management being a responsibility of the baseline `UiDraggable`... To me, the notion of a placeholder is an implementation detail of only one usecase of the drag'n'drop system - one that exists only to account for a quirk of layouts.  
It feels like this should be handled by a separate helper function/class that implements placeholders on top of a generic drag'n'drop framework.

scripts/mod_loader/ui/widgets/root.lua
Lemonymous: That's a very good question!
I was actually not considering buttons 2, 4 and 5 even existing. (I don't know why I forgot about button 3 though)

My suggested solution would be:
```lua
if not consumeEvent then
	consumeEvent = self:mousedown(mx, my, button)
end
```
right before
```lua
self:cleanupDropdown()
```
Would you agree?

scripts/mod_loader/ui/widgets/draggable.lua
Lemonymous: I can't see I completely see your issue, but I suppose I can see it potentially becoming one, if you envision drag-dropping being used in many other ways down the line.

So for a separate class, I can have this suggested solution. Would this be fine? (tested, and should be identical to the code in the PR except pulled out to a separate class)
```lua
UiDragDropList = Class.inherit(UiDraggable)

function UiDragDropList:new()
	UiDraggable.new(self)
end

function UiDragDropList:startDrag(mx, my, button)
	if button ~= 1 then
		return
	end

	UiDraggable.startDrag(self, mx, my, button)

	if self.dragPlaceholder ~= nil then
		local root = sdlext.getUiRoot()
		local owner = self.parent
		local index = list_indexof(owner.children, self)
		self:detach()
		self.owner = owner
		self.translucent = true

		owner:add(self.dragPlaceholder, index)
		self.dragPlaceholder:show()

		self:addTo(root.draggableUi)
		self.x = self.screenx
		self.y = self.screeny
	end
end

function UiDragDropList:stopDrag(mx, my, button)
	if button ~= 1 then
		return
	end

	if self.dragMoving and self.dragPlaceholder ~= nil then
		local index = list_indexof(self.owner.children, self.dragPlaceholder)
		self.dragPlaceholder:detach()
		self.dragPlaceholder:hide()

		self:detach()
		self:addTo(self.owner, index)
		self.translucent = self.dragPlaceholder.translucent
	end

	UiDraggable.stopDrag(self, mx, my, button)
end

function Ui:registerDragDropList(placeholder)
	Assert.NotEquals('nil', type(placeholder), "Argument #1: Must be an ui element")
	self:registerDragMove()
	self.dragPlaceholder = placeholder
	self.startDrag = UiDragDropList.startDrag
	self.stopDrag = UiDragDropList.stopDrag
end
```

Lemonymous: A sidenote: Maybe there should be an `Assert.IsClass` function?

scripts/mod_loader/ui/widgets/root.lua
kartoFlane: Yep, sounds good

scripts/mod_loader/ui/widgets/draggable.lua
kartoFlane: Yeah, that's class is pretty much what I had in mind.

Apart from a vague notion of what would make sense responsibility-wise, my main concern was with `stopDrag` reinserting the placeholder into the owner element, since with the `addTo` call, we depend on the target element's layout logic to correctly translate the element.

This works for elements that manage their children's positions (most/all of the `___Layout` classes), but breaks for elements that are just the bare `Ui` class (which is the reason we need to translate the placeholder when moving it to `root.draggableUi`). `UiDraggable` simply has insufficient information to account for that, since it doesn't really know how the parent element is going to behave, and when a correction is and isn't needed.

Regarding the assertions, I'm not completely against them, but we can already kinda do that via `Assert.True(Class.instanceOf(placeholder, Ui))`, and other cases via `isSuperclassOf`/`isSubclassOf`. I don't think having dedicated assertions for these would net much benefit, apart from being slightly shorter. Though these functions *could* use updating to not use base lua `assert`, to make them more informative in failure cases.

Lemonymous: I see. With that explanation I can certainly agree that the placeholder add/detach logic would not be the right fit for every type of parent element. Good call on moving it to another class.

Regarding the Asserts I was suggesting, I was not concerned with the longer code by using `Assert.True`. The reason I suggested `Assert.IsClass` (or a variation of it) was mostly to provide clearer error messages than `"[msg]: Expected true, but was false"`, which in itself doesn't say much of anything.

On further thinking though, classes don't know their own name, so it would in any case have been provided to the `Assert.IsClass` function. Using `Assert.True` I suppose maybe `"Argument#1 - IsSubClassOf(Ui): Expected true, but was false"` would be a sufficient solution for class checks.

scripts/mod_loader/ui/widgets/root.lua
```diff
@@ -97,66 +183,50 @@ function UiRoot:event(eventloop)
...
-		-- Notify pressed children of the event, even if the mouse is released
-		-- outside of them.
-		if self.pressedchild and self.pressedchild:mouseup(mx, my, button) then
-			self.pressedchild.pressed = false
-			self.pressedchild = nil
-			return true
+
+		if button == 1 and self.pressedchild ~= nil then
+			return self:releasePressedchild(mx, my, button)
+		elseif button == 3 and self.hoveredchild ~= nil then
```
Lemonymous, line 251: On further consideration; this elseif block is probably not the right call.
No system in the mod loader currently require to have their `mouseup` called on button 3 in order to consume the event before any of its ancestors have had their `mouseup` called.
I am considering removing it, so only mouse 1 will have a special case associated with it, and all other buttons (2-?) get their `mouseup` calls through the normal chain of `mouseup` calls via their ancestors. A `hoveredchild` should receive it as long as its ancestors haven't overridden their `mouseup` functions to prevent it. Which also doesn't seem unreasonable.

kartoFlane: Sounds good, it doesn't seem like it'll break anything as far as I can tell.

kartoFlane:
One last holdup I have before merging this is backwards compatibility with existing mods; mod loader's own UI seems to work fine, as do most UI mods that I have briefly checked. Two issues I've noticed:
* Into the Wild: fairly minor - in island arrange screen, dragging a CEO portrait causes the tooltip to be rendered at an incorrect position. *Probably* a matter of the way the UI is constructed in the mod itself?
* Hotseat: crashes - clicking End Turn as vek causes the game to crash due to `Ui:updateTooltipState` attempting to reference `root` field, which happens to be nil. It seems that some elements still receive clicks, even though they have been `detach`ed.

While the former I'd be willing to overlook, the latter seems like a bug in the way input events are sent to UI elements, and should be fixed before we can merge this.

Lemonymous:
`Into the Wild: fairly minor - in island arrange screen, dragging a CEO portrait causes the tooltip to be rendered at an incorrect position.`
Are you referring to how the tooltip will remain next to the cursor while dragging the element, instead of drawing the tooltip next to the element itself?
The current ui in master does not draw the tooltip at all if I recall. I changed it in this PR to keep drawing the tooltip while dragging; and added the function `Ui.updateTooltipState` to let each element decide how they want the tooltip to be drawn.
`pilot_arrange` and `palette_arrange` overrides `Ui.updateTooltipState` with the following:
```lua
local function updateTooltipState(self)
    self.root:settooltip(self.tooltip, self.tooltip_title, self.pressed)
end
```
One "solution" (if this should be default behavior) could be to add a `UiDragDropList.updateTooltipState`, so when an element uses `Ui.registerDragDropList`, the element is given this behvior. However, this wouldn't fix Into The Wild's ui, since it uses it's own logic for its drag-drop feature anyways.

The other issue is a bit deeper, so I'll have to get back to you on that.

Lemonymous:
Proposed fix for second issue:
```lua
if self.root then
   if self.root.focuschild == child then
      -- pass self as arg for UiRoot override
      self:setfocus(self)
   end

   if self.root.hoveredchild == child then
      self.root.hoveredchild = nil
      child.hovered = false
   end

   if self.root.pressedchild == child then
      self.root.pressedchild = nil
      child.pressed = false
   end
end
```
First check if the element even has a root, to allow orphaned elements to detach children.
Then pass focus to parent, and clear `hoveredchild` and `pressedchild` if applicable.
We could possibly do `self.root.updateHoveredState()` instead of clearing `hoveredchild` but I am not sure it's worth it. I _think_ it's better to keep it at a once per frame call.

Lemonymous:
After fixing the second issue, I am seeing additional bugs in Hotseat with how it assumes mouse events work. For instance, when deploying Vek spawns on the enemy turn, I have an element that expects `mousedown` to be called for a non-hovered element (`uiMain` in `phases\spawn.lua`).

(In this PR, if there is a hovered element when mouse button 1 is pressed; that element is the only one that received a `mousedown` call. On master, the call is also made to parents of the hovered element.)

Not completely sure how widespread this practice is. It might be isolated to Hotseat.

Lemonymous:
My previous proposed fix for issue 2 above is actually not sufficient for when Hotseat destroys its full enemy turn ui. At that point, root.hoveredchild and root.pressedchild may be pointing to an orphaned element.

New proposed fix:
```lua
function Ui:setroot(root)
    if self.root and self.root ~= root then
        if self.root.hoveredchild == self then
            self.root.hoveredchild = nil
            self.hovered = false
        end

        if self.root.pressedchild == self then
            self.root.pressedchild = nil
            self.pressed = false
        end
    end

    ...
```
`Ui.setroot` is called recursively on each child of the removed element.
If a new root is set that is different from current root, we can check if they are the hovered or pressed element; and update root accordingly.
Not sure on whether the focused element should have a similar treatment or not.

Lemonymous:
I'm seeing one additional issue with Hotseat.
A number of its ui elements uses a function `update` which is called when a resolution change is detected.

This PR creates a function with the same name in `ui/widgets/base.lua`; a function which is recursively called every frame for all elements under `UiRoot`. The intention of this new function was to let elements have a function they knew would be called, separate from `relayout`, responsible for further updating their state.

Whether the function is necessary, or useful, I don't know yet. However, it does conflict with Hotseat's non-standard way of updating its elements.

Some possible fixes could be:
- rename base Ui.update to `onUpdate`
- rename base Ui.update to `updateState`
- remove base Ui.update
- rename Hotseat's update functions

Lemonymous:
I now have a version of Hotseat that appears compatible with the changes proposed above.
However, I hold off, and listen to any input you have.

kartoFlane:
Huh, that turned out to run a little deeper than expected :P

> Are you referring to how the tooltip will remain next to the cursor while dragging the element, instead of drawing the tooltip next to the element itself?

At least on my PC (`rework-ui-system` brach, fullscreen), the tooltip is drawn completely away from the intended element. The tooltip is positioned relative to top-left corner of the screen, rather than top-left corner of the arrange islands window - the offset itself appears correct, just applied relative to incorrect element.

> However, this wouldn't fix Into The Wild's ui, since it uses it's own logic for its drag-drop feature anyways.

If that's the case, then it's probably something that needs to be fixed in ITW, since the tooltips appear to work properly for other UI elements that rely on mod loader's solution.

&nbsp;

> (In this PR, if there is a hovered element when mouse button 1 is pressed; that element is the only one that received a mousedown call. On master, the call is also made to parents of the hovered element, iirc.)

I think I found the problem: `Ui:handleMouseEvent` always returns `true`, so hovered elements always intercept all mouse events, even if they don't do anything with them. Seems like `base.lua:281` should be changed to `return false`

> If a new root is set that is different from current root, we can check if they are the hovered or pressed element; and update root accordingly.
> Not sure on whether the focused element should have a similar treatment or not.

Doesn't really make sense for focus to remain on an orphaned element, since we'll have the same problem as this, but for keyboard events. I think focus should be updated as well.

> rename base `Ui.update` to `Ui.updateState`

I think we want to keep this function around. `updateState` isn't a bad name anyway, so we can go with that.  
While we're at it, I think it would be a good idea to separate animation updates from `draw` into an `updateAnimations` function, invoked with the others - IIRC you recently added a workaround for that that involved temporarily zeroing `deltaTime`, so that hack wouldn't be required anymore.

kartoFlane:
> I'm not 100% sure of the intent behind `root.focuschild` and `ui.focused` though. My first assumption is that only one element can be focused at one time; but the method `UiRoot:setfocus` changes focused for many elements, not just one. I don't know if this is for redundancy or another reason.

`root.focuschild` is the element that has focus, and should logically receive events, etc. `ui.focused` is a marker for decorations, so that they draw the elements, or parent containers, with correct style. I don't think this is currently being used anywhere.  
Frankly, the focus 'system' isn't very developed or particularly well thought out, so there's definitely some room for improvement.

> It indeed does clear `hoveredchild` and `pressedchild` whenever elements are detached, so there are no errors thrown. However, our ui drag drop systems, remove and reattach elements when we drag them, so we can move them around freely. For a moment after removing the element, it is orphaned, so `hoveredchild`/`pressedchild` is cleared, which results in us not being able to drag the element at all, since it is released the moment we start dragging.
> I don't have any solution for this yet.

I think in this case, the code that implements the drag'n'drop system should handle updating `hoveredchild` and other variables after moving the element between parents. There isn't really another way around it, I think.

Lemonymous:
> root.focuschild is the element that has focus, and should logically receive events, etc. ui.focused is a marker for decorations, so that they draw the elements, or parent containers, with correct style. I don't think this is currently being used anywhere.
> Frankly, the focus 'system' isn't very developed or particularly well thought out, so there's definitely some room for improvement.
Alright, that is what I thought it did. I was just a bit confused by `UiRoot:setfocus` thoroughness in clearing the `focused` tag from ui elements.
It might not be very developed, but it does seem important for which elements receive keyboard events.
> I think in this case, the code that implements the drag'n'drop system should handle updating hoveredchild and other variables after moving the element between parents. There isn't really another way around it, I think.
I guess that might be the cleanest solution. I'll keep thinking about it for a bit longer though.

kartoFlane:
So it looks like we've addressed everthing. Anything else left to add/fix on your end?

Lemonymous:
I thought a little about the `mousedown` event only going to the hovered element.
- On one hand, I think it is a good idea; it's fast and we know the event goes to the hovered element no matter what (no element can block the signal from passing through).
- On the other, it is a change from the original system. It also diverges from the standard behavior of all other events (bar `mouseup`, which is only passed to the pressed element, since we don't know whether that one contains the mouse or not)

Seeing as `containsMouse` is cached, passing events down from parent to children shouldn't cost all that much. I'm just wondering if symmetry of how the events operate could maybe make them easier to use. I don't know how much that is worth though.

That being said, I don't have any more code to push to this PR. It all seems to be working, so I would be fine with it being like it is.

PR #79, opened 2021-04-20
Feature UI input

Lemonymous:
**NOTE**: this PR is now reliant on `Github.com/Lemonymous/IntoTheBreachLua/tree/fix-text-surfaces`
**NOTE**: this PR is now reliant on `Github.com/Lemonymous/IntoTheBreachLua/tree/feature-clipboard`

Using the sdl event added by kartoFlane/IntoTheBreachLua#6.

Adds an event that is fired when a keyboard key which produces a readable character is pressed.

### New event: `onTextInput`
`modApi.events.onTextInput`
Event is dispatched to every subscriber.

Example use:
```lua
modApi.events.onTextInput:subscribe(function(textInput)
    LOG("Pressed character ".. textInput)
    return true
end
```
### New method `Ui.textInput`
`Ui.textInput(textInput)`
Called for the focused element.

Example use:
```lua
function ui:textInput(textInput)
    LOG("Pressed character ".. textInput)
    return true
end
```

### New Ui Class `UiTextBox`
`UiTextBox` can either be used to create ui elements of Class `UiTextBox`, or to apply its functions to another Ui Class instance with `Ui.registerAsTextBox`

#### This class has the fields:
- `caret` - the position of the caret which is the entry point for adding and removing text.
- `typedtext` - the text of the widget.
- `editable` - if set to false, any method that adds/removes text will return without doing anything.
- `selection` - if non-nil: a selection exists between this number and `caret`.

#### The following methods are used to edit `typedtext`:
- `addText` - adds the specified text at the caret position.
- `delete` - deletes a specified amount of text behind the caret position.
- `backspace` - deletes a specified amount of text in front of the caret position.
- `newline` - adds a linebreak at the caret position.
- `onInput` - called on input events with the widget focused. Defaults to add the input text at the caret position. If a selection exists, the selection will be deleted first.
- `onEnter` - called when pressing enter with the widget focuesd. Defaults to call `newline`. If a selection exists, the selection will be deleted first.
- `onDelete` - called when pressing delete with the widget focused. Defaults to delete 1 character behind the caret position. If a selection exists, the selection will be deleted instead.
- `onBackspace` - called when pressing backspace with the widget focused. Defaults to delete 1 character in front of the caret position. If a selection exists, the selection will be deleted instead.

#### The following methods changes the caret position and selection:
- `setCaret` - sets the caret to the specified position
- `moveCaret` - moves the caret position forwards or backwards a specified amount
- `onArrowRight` - called when pressing arrow right with the widget focused. Defaults to moving the caret position 1 right, or if ctrl is held, 1 word to the right. If shift is held, text will be selected.
- `onArrowLeft` - called when pressing arrow left with the widget focused. Defaults to moving the caret position 1 left, or if ctrl is held, 1 word to the left. If shift is held, text will be selected.
- `onArrowUp` - called when pressing arrow up with the widget focused. No default behavior.
- `onArrowDown` - called when pressing arrow down with the widget focused. No default behavior.
- `onHome` - called when pressing arrow down with the widget focused. Defaults to moving the caret to the very beginning. If shift is held, text will be selected.
- `onEnd` - called when pressing arrow down with the widget focused. Defaults to moving the caret to the very end. If shift is held, text will be selected.
- `onPageUp` - called when pressing arrow down with the widget focused. No default behavior.
- `onPageDown` - called when pressing arrow down with the widget focused. No default behavior.
- `onSelectAll` - called when pressing ctrl+a with the widget focused. Defaults to selecting the entire text.

#### The following methods handles copy/paste functionality:
- `copy` - copies the selected area to the clipboard.
- `paste` - pastes the clipboard text contents into the widget at the caret position.
- `onCopy` - called when pressing ctrl+c with the widget focused. Defaults to copying the selected text to the clipboard.
- `onCut` - called when pressing ctrl+x with the widget focused. Defaults to copying the selected text to the clipboard, and deleting the selection from the widget.
- `onPaste` - called when pressing ctrl+v with the widget focused. Defaults to pasting the clipboard text contents into the widget.

### New decoration `DecoTextBox`
`DecoTextBox` is a decoration tailored to `UiTextBox`, made to dynamically be updated whenever the text changes.
The decoration has many options in how to display the text, but is mainly intended to be kept inside the container ui element it decorates.

#### Methods:
- `setPrefix` - sets a fixed text to the start of the displayed text.
- `setSuffix` - sets a fixed text to the end of the displayed text.
- `setWrapText` - enable or disable text wrapping. Text wrapping will attempt to contain the text within the left and right edges of the widget.
- `setSplitWords` - enable or disable word splitting. Word splitting together with text wrapping will prefer to split text as soon as necessary, instead of trying to keep each word on a single line.
- `setTextSettings` - sets the text settings used to draw text.
- `setFont` - sets the font used to draw text.
- `caretToScreen` - converts a caret position to screen coordinates.
- `screenToCaret` - converts a screen coordinate to a caret position.
- `drawCaret` - draws the caret at the widget's caret position.
- `onArrowUp` - the widget's onArrowUp method is replaced and redirected to this method, which will convert an arrow up press to moving the caret position one line up. If shift is held, text will be selected.
- `onArrowDown` - the widget's onArrowDown method is replaced and redirected to this method, which will convert an arrow down press to moving the caret position one line down. If shift is held, text will be selected.
- `mousedown` - the widget's mousedown method is replaced and redirected to this method; which will convert a mouse press to moving the caret position to the position closest to the mouse cursor. A selection will be initiated, which can be continued by dragging the mouse cursor.
- `mousemove` - the widget's mousemove method is replaced and redirected to this method; which will convert a mouse drag to updating the selection and the caret position to the position closest to the mouse cursor.


kartoFlane:
What's the status of `fix-text-surfaces` branch? This PR relies on it, and I see that branch hasn't been submitted as a PR yet, but the commit on it seems worded as if it is

Lemonymous:
I'm not 100% sure what happened there to be honest. The commit certainly looks like I intended to submit it as a PR. I will look into it.

As for this PR. I do not believe it is finished.

After you made me aware of your branch https://github.com/kartoFlane/ITB-ModLoader/tree/feature/ui-textfield-widget, I've wanted to add more features to it to make it a proper textbox with selections, copy/cut/paste, arbitrary caret placement, ctrl+arrow/backspace to skip/erase whole words, home/end/pageup/pagedown support, etc.

I also want to change the logic so only the text between newlines are worked on internally, to improve performance on texts with multiple paragraphs; since text before and after newline in relation to the current caret placement should remain the same, as long as the textbox dimensions and font are not altered.

I just haven't gotten that far yet.

scripts/mod_loader/ui/decorations/deco_textbox.lua
```diff
@@ -0,0 +1,657 @@
...
+	-- Potential performance gain:
+	-- only calculate characters inside the widget. This is currently
+	-- very difficult, due to every character's position being
+	-- determined from start to end at drawtime
+	
+	-- fill drawbuffer
+	for i = 1, textLength do
+		local char = getCharacter(textfield, i, 1)
+		commit(wrap, char, i)
+	end
```
kartoFlane, line 523: If my scuffed time-measuring code is to be trusted, this is the most performance-critical section - it takes up over 90% of total time spent in the `draw()` function (with ~ 27k characters in the textbox). Actually drawing the surfaces takes "only" 10% of the time. The rest of the method seems to have no significant performance impact.

kartoFlane: Simply wrapping the drawbuffer update logic in an `if` block allows to maintain capped FPS for 27k chars, with *relatively minor* dips (ie. halved FPS) when the buffer needs to be updated, eg.:

```lua
if self.dirtybuffer then
        -- fill drawbuffer
        ...

	self.dirtybuffer = false
end
```
but that causes scrolling and selection to break, I'm not sure what exactly relies on the buffer, and needs to be updated along with it.

Lemonymous:
This PR can probably be closed. #145 adds an input field with a slightly smaller scope.

PR #81, opened 2021-04-25
Various minor fixes

Lemonymous:
Various minor fixes and changes

### DIFF
--- a/scripts/mod_loader/modapi/board.lua
+++ b/scripts/mod_loader/modapi/board.lua
@@ -50,7 +50,7 @@ BoardClass.SetShield = function(self, loc, shield)
 	Assert.Equals("boolean", type(shield), "Argument #2")
 
 	local dmg = SpaceDamage(loc)
-	dmg.iShield = shield and EFFECT_CREATE or EFFECT_REMOVE
+	dmg.iShield = shield and EFFECT_CREATE or -1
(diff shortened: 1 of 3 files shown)


scripts/mod_loader/modapi/board.lua
```diff
@@ -50,7 +50,7 @@ BoardClass.SetShield = function(self, loc, shield)
 	Assert.Equals("boolean", type(shield), "Argument #2")
 
 	local dmg = SpaceDamage(loc)
-	dmg.iShield = shield and EFFECT_CREATE or EFFECT_REMOVE
+	dmg.iShield = shield and EFFECT_CREATE or -1
```
kartoFlane, line 53: Interesting, seems that the `EFFECT_REMOVE` global returns a value of `2`... I'm pretty sure it used to be `-1`

Lemonymous: I suppose it is possible it was -1 at some point. I don't remember it ever was though. I still have a 1.1.22 GoG version around, and it is 2 there.

The various effects (fire/frozen/acid/shield/smoke) have always required slightly different considerations when adding and removing them. Acid for instance, can be created with SpaceDamage iAcid = 1, but cannot be removed with SpaceDamage at all.

Issue #83, opened 2021-05-25
GetCurrentMission can return a mission after the mission has been completed


Lemonymous:
`GetCurrentMission` will return a mission table even when there is no current mission by executing the following steps:
- win a mission.
- while the mission has been completed, but not yet returned to the island view, exit to the main menu.
- load the game back up to return to the mission's end screen. This seems skip past the code that would clear the current mission variable (by design it appears in the code).
- `GetCurrentMission` will now be in a state where it will return this mission until either a new mission has been started, or exiting to the main menu again.

If the game is not exited while the mission is in the process of ending; `GetCurrentMission` will stop returning the current mission when the board stops being `busy`. It looks like this is by design, but I don't know for what reason.

RegionData.iBattleRegion appears to reflect the current mission more accurately than the current implementation of `GetCurrentMission`. RegionData.iBattleRegion does however, keep returning the mission for the full duration of the end mission screen, and until the player gains control to select a new mission on the current island. This is also the exact point in time where exiting to the main menu and returning to the game, will no longer return to the end mission screen for that mission.

kartoFlane:
Resolved in #84

PR #84, opened 2021-05-25
Fix current mission

Lemonymous:
Referencing issue #83

This is a proposed change to `GetCurrentMission`

The reason for wanting to fix `GetCurrentMission`, is that I wanted to create some functionality that relied on detecting the current mission, and the current implementation is not accurate enough for my purposes.

However, since this is an old function; and a lot of code may rely on the current implementation, I am open to it needing to remain unchanged.

### DIFF
--- a/scripts/mod_loader/altered/misc.lua
+++ b/scripts/mod_loader/altered/misc.lua
@@ -113,14 +113,6 @@ local function overrideNextPhase()
 	end
 end
 
-local function updateCurrentMission()
-	local region = GetCurrentRegion()
-	
-	if region then
-		modApi.current_mission = GAME:GetMission(region.mission)
-	end
-end
-
 local oldStartNewGame = startNewGame
 function startNewGame()
 
@@ -186,7 +178,6 @@ function LoadGame()
 
 	RestoreGameVariables(Settings)
 	overrideNextPhase()
-	updateCurrentMission()
 
 	if GAME.CustomDifficulty then
 		SetDifficulty(GAME.CustomDifficulty)
--- a/scripts/mod_loader/altered/misc.lua
+++ b/scripts/mod_loader/altered/misc.lua
@@ -105,7 +105,8 @@ local function overrideNextPhase()
 
 		if nxtId ~= "" then
 			-- Set the mission's id, since the game doesn't do it
-			local nextMission = _G[nxtId]
+			local id = self:GetMissionId(mission)
+			local nextMission = self.Missions[id]
 			nextMission.ID = nxtId
 
 			modApi:fireMissionNextPhaseCreatedHooks(prevMission, nextMission)
--- a/scripts/mod_loader/altered/missions.lua
+++ b/scripts/mod_loader/altered/missions.lua
@@ -67,6 +67,7 @@ end
 
 local oldBaseDeployment = Mission.BaseDeployment
 function Mission:BaseDeployment()
+	modApi:setMission(self)
 	oldBaseDeployment(self)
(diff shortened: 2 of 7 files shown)


kartoFlane:
> If the game is not exited while the mission is in the process of ending; `GetCurrentMission` will stop returning the current mission when the board stops being `busy`. It looks like this is by design, but I don't know for what reason.

The reason for that is so that `modApi.current_mission` and `modApi.runLaterQueue` aren't cleared while mission-end effects are still happening on the board. I'm not entirely sure if waiting for the board to not be busy anymore is required to avoid all problematic edge-cases. Maybe executing that in a regular `AddScript()` like you suggest is enough /shrug

If I understand correctly, it seems like `runLaterQueue` is not cleared anywhere else, and so the queued action happens once you return to the mission from main menu? I'm pretty sure I added code specifically to avoid that, but if not, that's an oversight - we don't want things to persist in `runLaterQueue` when going to main menu and back, since attempting to accommodate actions to stick around between reloads like that sounds like a headache for everyone involved (modders included).

I think it's better to just treat going to main menu and back the same as restarting the game completely, so that both instances can be handled using the same approach, ie. assuming a blank slate and that everything needs to be restored from scratch.

Lemonymous:
> >     If the game is not exited while the mission is in the process of ending; GetCurrentMission will stop returning the current mission when the board stops being busy. It looks like this is by design, but I don't know for what reason.

>
> The reason for that is so that modApi.current_mission and modApi.runLaterQueue aren't cleared while mission-end effects are still happening on the board. I'm not entirely sure if waiting for the board to not be busy anymore is required to avoid all problematic edge-cases. Maybe executing that in a regular AddScript() like you suggest is enough /shrug

I may have worded that badly. Current code has a conditional hook, clearing `modApi.current_mission` if returning to the menu, or if the board stops being busy. I believe `modApi.current_mission` gets set again after loading the game, since `Region.iBattleRegion` still points to that particular mission for the duration of the end mission screen.
I am asking if there is a reason for clearing `modApi.currentmission` at the point that it is (non-busy board); not as opposed to clearing it immediately; but as opposed to clearing it when `RegionData.iBattleRegion` returns nil, when the mission is completely over in every sense.

> If I understand correctly, it seems like runLaterQueue is not cleared anywhere else, and so the queued action happens once you return to the mission from main menu?

According to the code, it appears the `runLaterQueue` is currently cleared immediately as the mission ends, and the conditional hook is also cleared when exiting the game.

> I think it's better to just treat going to main menu and back the same as restarting the game completely, so that both instances can be handled using the same approach, ie. assuming a blank slate and that everything needs to be restored from scratch.

I would agree with this.
If my proposal appear to go against this, please point it out. I am not sure I see it if that is the case.

Lemonymous:
`GetCurrentMission` would not return the current mission in an `BaseDeployment` function due to `RegionData.iBattleRegion` not being set at that time. This new code should address this issue. It also goes back to the idea of caching the current mission instead of always fetching it.

By updating the current mission every time the save data changes, it should hopefully stay up to date in all situations.
Initial tests using the `onMissionChanged` event looks promising.

Test code:
```lua
modApi.events.onMissionChanged:subscribe(function(newMission, oldMission)
	LOGF("mission changed from %s to %s", tostring(oldMission), tostring(newMission))
end)
```

kartoFlane:
Actually, it looks like I misunderstood the problem at first... I don't see anything wrong with the changes here right now.

Lemonymous:
Great!
I have nothing further to add to this PR.

PR #92, opened 2021-07-20
Added missing parameters to onGameVictory event dispatch

kartoFlane:
Parameter values taken from medal-awarding code


Lemonymous:
I took a look at the code I wrote, and your changes does indeed make a lot more sense.
It looks like I had the same idea, but somehow failed to complete it.

PR #103, opened 2021-07-27
Create mission.QueuedSpawns if it is nil in Mission:UpdateQueuedSpawns

Lemonymous:


Lemonymous:
After doing some more investigating; weird as this seems, calling the following code when testing a map in mapedit will produce the following error.

This PR alleviates this issue.

Code:
```lua
local fx = SkillEffect() 
fx:AddDelay(1) 
Board:AddEffect(fx)
```
Error:
`./scripts/mod_loader/altered/missions.lua:21: attempt to get length of field 'QueuedSpawns' (a nil value)`

kartoFlane:
Right, I only set the variable in `Mission:BaseStart`, but that doesn't run in mapedit... Maybe we could invoke `Mission:BaseStart` once we enter mapedit, if we can detect it somehow.

Lemonymous:
> Maybe we could invoke Mission:BaseStart once we enter mapedit

Could make an issue about it as an enhancement.
I don't think invoking BaseStart should be done without a lot of testing though, as testing a map in mapedit is very different from starting an actual mission. The conditions are vastly different.

Right now, this fix solves this immediate issue, and certainly does no harm.

Issue #104, opened 2021-07-28
mapedit has not been taken much into account, and can crash; especially with mods.


Lemonymous:
The console command `mapedit` will bring up a map editor. Maps in the map editor can be tested, which launches a non-standard mission with a pretty bare-bones mission object. This mission object is the same for every mission in the map editor.

If we can find this mission object, and can detect when this mission starts and ends, we could try to make it operate more similar to a standard mission. Or alternatively, maybe make it operate more similar to Test Mech Scenario, since both of these missions share the trait of not have a save game to refer to.

I have spoken briefly with narD, who has made several videos showcasing neat interactions in "narD's Lab" using `mapedit`. They could confirm that the map editor is not stable when using it with mods.

Lemonymous:
Update:
The mission object used in `mapedit` is in fact the same one used in Test Mech Scenario: `_G["Mission_Test"]`

PR #105, opened 2021-07-28
Added events for when entering/exiting map editor maps

Lemonymous:
This is reusing windows detection code for simplicity, even though the map editor map is technically not a window.

Referencing #104


Lemonymous:
`Mission_Test` is used for both Test Mech Scenario and Map Editor Scenario.

Since some of the functions in `Mission_Test` refers specifically to Test Mech Scenario, some more work will have to be done to correctly separate the two.

I will not add that code to this PR.

PR #113, opened 2021-08-06
Added onPawnSelected and onPawnDeselected events

Lemonymous:
Events are added into the same loop that handles `onPawnFocused` and `onPawnUnfocused` events, because the loop already fetches the information about the selected pawn every frame.

Due to how exiting a run to the main menu causes both `Game` and `Board` to seize to exist, `onPawnDeselected` will not be dispatched if doing so while a pawn is selected.

The alternative was to dispatch with the pawn id as argument instead. Both variants have some merit.

The current way will require the user to subscribe to an `onGameExited` event to free up anything if `onPawnSelected` and `onPawnDeselected` are working in tandem.

### DIFF
--- a/scripts/mod_loader/altered/skills.lua
+++ b/scripts/mod_loader/altered/skills.lua
@@ -176,6 +176,16 @@ end
 modApi.events.onFrameDrawStart:subscribe(function()
 	local selectedPawnIdCurrent = Board and Board:GetSelectedPawnId() or nil
 
+	if Game and selectedPawnIdCurrent ~= selectedPawnIdLast then
+		if selectedPawnIdLast then
+			modApi:firePawnDeselectedHooks(Game:GetPawn(selectedPawnIdLast))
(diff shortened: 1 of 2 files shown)


kartoFlane:
> The current way will require the user to subscribe to an `onGameExited` event to free up anything if `onPawnSelected` and `onPawnDeselected` are working in tandem.

That's how it will have to be, I think - even if we send the pawn's id as argument and change `onPawnDeselected` to be dispatched on game exit, modders will probably try to fetch the Pawn from `Game` and then scratch their heads why their hook fails - then they'll have to add `if Game then` to their hook. I think handling this explicitly in `onGameExited` is a better approach overall, since that's what they'd have to do anyway.

Lemonymous:
Yes, that was my conclusion as well after some forth and back.
It's not a perfect solution, but I think it is the least bad we have at the moment.

PR #129, opened 2022-07-20
[RFC] Remove custom difficulty

KnightMiner:
Previously, the mod loaded added in two higher difficulties, "Very Hard" and "Impossible". It was left up to mods to implement these difficulties, and since few did it often left players confused and added one more thing to port each update.

Advanced Edition added in the new Unfair difficulty which makes the game even more unforgiving than hard mode, and serves the same purpose as "Very Hard" and "Impossible". As a result, I am proposing we remove the mostly unused custom difficulty feature in favor of the new Unfair difficultly.

Moving forwards, mods wishing to add custom difficultly can do so using config options similar to the Advanced Edition content toggles.

----

Notes:
* This PR removes any hooks that are no longer functional, such as `SetDifficulty` and `AddDifficultyLevel`
* Any functions that we can provide a reasonable implementation to were kept in `compat`
* The function `Mission:SetupDifficulty` was kept, allowing mods to use that to setup their difficulty based on their mod config.
* I have tested this change by "completing" a full run using cheats, and did not see any errors in either log.


kartoFlane:
I agree - I also ended up regretting adding this functionality as part of the mod loader. Also, the config options specific to each mod ended up being much more configurable and useful than the difficulty selection, which has always been buggy.

KnightMiner:
Going to go ahead and merge this so we can get a pre-release out.

Modders: if you have concerns about this change, please make them known before a full modloader release

PR #132, opened 2022-07-23
Improve efficiency of modifying resource.dat file

KnightMiner:
Did a loose benchmark:

* Old startup time:
    * With no resource.dat.bak: around 50 seconds
    * With resource.dat.bak: around 65 seconds
* After changes
    * With no resource.dat.bak: around 27 seconds
    * With resource.dat.bak: around 30 seconds

Breakdown of improvement by change:
* Switching from list search to hash search: 3-4 seconds
* Removing stream copying: 14 seconds with no resource.dat.bak, around 30 seconds with it
* Faster signature scan: around 3 seconds

Note: startup times include the loading bar, which takes about 7 seconds. Meaning all the modloader preloading stuff takes around 40-55 seconds before changes, and around 20-23 seconds after.

----

Note: we could possibly save a few more seconds if we inject our signature into the start of `resource.dat` and break the signature scan loop as soon as its found. Based on how fast it ran with the broken signature scan code I had earlier, I doubt it would be much more than a second gain, so probably not worth it (plus, not sure if there is risk with changing the order of files)

----

Testing:
* If you contribute to the modloader, you know how to test
* If not, test by installing the mod loader using this link: [ITB-ModLoader-2.7.0-faster-resource-dat.zip](https://github.com/itb-community/ITB-ModLoader/files/9174852/ITB-ModLoader-2.7.0-faster-resource-dat.zip)

### DIFF
--- a/scripts/mod_loader/ftldat/ftldat.lua
+++ b/scripts/mod_loader/ftldat/ftldat.lua
@@ -21,23 +21,23 @@ function FtlDat:_init(p__io, p__parent, p__root)
 	self.signature = false
 end
 
-function FtlDat:_read()
+function FtlDat:_read(signature_scan)
 	self._numFiles = self._io:read_u4le()
 
 	self._files = {}
 	self._filesByName = {}
 	for i = 1, self._numFiles do
 		local file = File(self._io, self, self.m_root)
-		file:_read()
+		file:_read(signature_scan)
 		self:_insert_file(file)
 	end
 end
 
 function FtlDat:_write()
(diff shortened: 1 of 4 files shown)


Lemonymous:
I have not gone through the code, but I did test it on my system. On my system, it cuts the loading of the mod loader without mods to a third of `master`.

KnightMiner:
Not sure what could cause that apart from a mod manually modifying the resource.dat, there is no path where `assertExists` would have worked with the old code where it does not work with the new code

PR #133, opened 2022-07-24
Add compatibility for FURL library & fix the palette offset error it introduces

kartoFlane:
It seems the palette offset error happens due to some weird interaction between two versions of the FURL library; an older one, and a newer one that can also handle pilots -- at least, that's the difference between the two mods that are guaranteed to reproduce the issue (Treeherders (old FURL) and Weapons of the Past (new FURL)) -- though curiously, in both versions of the library, the code that is executed when adding palettes is the same.

This PR fixes the issue by altering part of FURL library's code before mods load it, and redirecting it to the mod loader's compatibility code, which in turn uses mod loader's existing palette API. To achieve this, a new event `modApi.events.onModEnumerating` has been added, which takes the path to the mod's directory as argument, and runs before the mod's `init.lua` file is executed.

Detection of whether a mod uses FURL or not is very naive - we simply check if the mod contains a file named `FURL.lua` in the root of its scripts directory. While mods could have it in different path, all of the mods I checked had it in that standard location, so it should be good enough for now. We can add more complex logic if we need to.

## Testing

The bug is pretty elusive, but the most consistent way to reproduce it we've found so far is by having both Treeherders and Weapons of the Past enabled and loaded at the same time.


Lemonymous:
While this PR probably fixes the error, the theory doesn't seem to be correct.

Made 1 mod with just old FURL + 1 color and 1 mod with just new FURL and 1 color -> no shift
Treeherders + simple mod with old FURL w/1 color -> no shift
Treeherders + simple mod with new FURL w/1 color -> no shift
Treeherders + Eclipse (does not run FURL, but customPalettes instead) -> shift

Lemonymous:
Now that this issue is fixed in fb5628f, I think this PR should be closed. Replacing file content of mods should probably be avoided.

KnightMiner:
The fix I applied is only relevant to the advanced edition content. Before AE, the only paletted images were at `units/player`, now `advanced/units/player` is also paletted.

It might be useful to have a way to altar mods if needed like we have a way to altar the base game scripts, but since the FURL bug is fixed we could leave that part out

kartoFlane:
Considering that the main purpose of this PR is no longer valid, I'm closing it and moving the `onModEnumerating` functionality to #137

PR #134, opened 2022-07-26
Improve `modApi:getGameVersion`

kartoFlane:
This PR changes the `modApi:getGameVersion` function so that it reads the version from the game's own logs at startup, rather than attempting to guess based on what functions are available.


Lemonymous:
Seems to work to return the current game version.

Tested:
- Deleted all savedata - required starting ITB twice (as normal), and version is correct.
- Deleted log.txt - game recreates it on startup before it is read.

PR #137, opened 2022-07-28
Add `onModEnumerating` event

kartoFlane:
This PR adds a new event `modApi.events.onModEnumerating`, which takes the path to a mod's directory as argument, and runs before the mod's init.lua file is executed. This allows the mod loader to alter mod files before they're loaded to apply fixes, or inspect mod content.


kartoFlane:
I'm closing this PR, seeing as currently there isn't really a use case for this event.

PR #145, opened 2022-08-04
Feature input field

Lemonymous:
## New
- `UiInputField` widget.
- `DecoInputField` decoration.
- `SurfaceFont` object - takes font and text settings, and returns an object that lazily creates surfaces for single characters.
- General `BinarySearch` function.
- Tests to the mod loader testing console for `BinarySearch` function.

Example input field:
```lua
local inputField = UiInputField()
  :widthpx(200):heightpx(40)
  :clip()
  :decorate{
	  DecoInputField{
		  alignV = "center",
		  offsetX = 5,
	  },
	  DecoBorder(),
  }
```


scripts/mod_loader/bootstrap/binarySearch.lua
```diff
@@ -0,0 +1,32 @@
+
+-- Find the element with the smallest return, larger than target.
+-- Assumes a list with sorted returns in ascending order.
+function BinarySearchMin(left, right, target, compareFn)
```
KnightMiner, line 4: This is a pretty non-standard compare function, standard ones are of the form `function(a, b)` which returns negative, `0`, or positive.

could you either add some docs explaining how it works or reply in the comment here?

Lemonymous: Would something like this be better?
```lua
-- Given a continuous ascending sorted list between index `left` and index `right`;
-- Find the smallest index where the predicate `isSmaller(index)` returns false.
function BinarySearchMin(left, right, isSmaller)
	local floor = math.floor
	while left < right do
		local middle = floor((left + right) / 2)
		if isSmaller(middle) then
			left = middle + 1
		else
			right = middle
		end
	end

	return left
end
```

KnightMiner: That does seem a bit easier to use.

Since binary aearches are typically though of using an array, it might be helpful to make the example show how you would use this on an array. I believe you would pass

```lua
BinarySearchMin(1, #array, function(index) return array[index] < value)
```
So before was basically an array indexing function if I understand this

Lemonymous: Yes, that example would be the correct use.

I do agree it is quite non-standard though. I made the function not take an array as input basically because the buffer in DecoInputField is a 0-indexed array containing alternating entries of surfaces and x draw position.

Maybe the better solution is to just move the function BinarySearchMin into DecoInputField.

KnightMiner: Worth noting, `BinarySearchMin` and `BinarySearchMax` are quite similar. Since the condition differnce is combined into the new `isSmaller` function, all that remains is how `left` and `right` update. Could potentially merge them as `BinarySearchIndex` with a parameter to specify if equal is left or right (that is, does right subtract from the middle or left add) that is a name unlikely to conflict later. 

I think the only concern I have with it being global is someone may make a global binary search later that works with arrays, functions in the global namespace should be relatively universal. That said, we could always leave the functions as is and make them local. Its far easier to make a local function global than to make a global function local once its in a release.

PR #152, opened 2022-08-19
Sort squad units when adding squad

Lemonymous:
The game orders units in a specific way in the hangar. Sort modded squads in the same order.
This should fix the issue where some modded squads would not display their squad achievements in the hangar ui.

### DIFF
--- a/scripts/mod_loader/modapi/squad.lua
+++ b/scripts/mod_loader/modapi/squad.lua
@@ -37,6 +37,9 @@ function modApi:addSquad(squad, name, desc, icon)
 	squad.id = squad.id or buildSquadId(squad)
 	Assert.Equals(nil, modApi.mod_squads_by_id[squad.id], string.format("Squad with id %q already exists", squad.id))
 
+	-- sort the squad name first
+	local mechOrder = {[squad[1]] = -1}
+
 	-- Validate the squad
 	for i = 2, 4 do
 		local mechType = squad[i]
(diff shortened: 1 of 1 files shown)


scripts/mod_loader/modapi/squad.lua
```diff
@@ -43,6 +51,17 @@ function modApi:addSquad(squad, name, desc, icon)
...
+		if i > 2 then
+			local ptable_prev = _G[squad[i-1]]
+			local pri = CLASS_ORDER[ptable.Class] or INT_MAX
+			local pri_prev = CLASS_ORDER[ptable_prev.Class] or INT_MAX
+
+			if pri < pri_prev then
+				-- swap entries
+				squad[i-1], squad[i] = squad[i], squad[i-1]
+			end
+		end
```
KnightMiner, line 64: This will not work in the case of `{"SquadName", "Brute", "Ranged", "Prime"}`, that is where the last element belongs at the start of the list. You may be better off doing something like the following:
```lua
local mechOrder = {[squad[1]] = -1} -- sort the squad name first
for i = 2, 4 do
  		local mechType = squad[i]
		local ptable = _G[mechType]
        -- validate
        mechOrder[mechType] = CLASS_ORDER[ptable.Class] or INT_MAX
end
-- sort the mechs based on the squad order
table.sort(squad, function(a, b) mechOrder[a] < mechOrder[b])
```
I am pretty sure this will preserve the order of duplicate mechs of the same class, but that should be tested

PR #153, opened 2022-08-20
Rework how skill effect metadata is stored to fix #140

### LINKED ISSUE #140: [2.7.1] Cannot earn third Bombermechs achievement with modloader
Powered Blast (Pierce a walking bomb with the AP cannon to kill an enemy) cannot be earned with the modloader. Was tested with no mods installed and still did not work.

KnightMiner:
The third bombermech achievement relies on the second space damage in the skill effect being a killing blow, and the third skill effect targeting a bomberling. However, since the second space damage in the skill effect is a projectile, the modloader injects metadata before that effect in the final return.

To fix this, rework the skill effect so that instead of storing the metadata as a space damage, it is stored in a separate table on the skill effect object. Due to how userdata works, we cannot store the metadata on the space damage or the damage list as neither persists

To reduce the burden of copying skill effects (and to make modders lives easier), two new functions are added: `DamageList:AppendAll()` and `SkillEffect:AppendAll()` which copy all space damage from one skill to another.

### Breaking changes

This PR will be a breaking change for anyone currently copying skill effects, as they will lose track of metadata. Additionally, this will break usages of `SkillEffect:GetMetadata()`, `SkillEfffect:GetQueuedMetadata()`, and `DamageList:GetMetadata()`. Since the size of the damage list is no longer the same, there was no way to avoid that breaking change

### DIFF
--- a/scripts/mod_loader/modapi/skills.lua
+++ b/scripts/mod_loader/modapi/skills.lua
@@ -55,41 +55,81 @@ function SpaceDamage:ToTable()
 	return result
 end
 
+-- deprecated: space damage is no longer metadata
 function SpaceDamage:IsMetadata()
-    Assert.Equals("userdata", type(self), "Argument #0")
-
-    if not self.sScript then
-        return false
-    end
-    
-    return modApi:stringStartsWith(self.sScript, "return")
+    return false
 end
 
-function DamageList:GetMetadata()
-    local metadata = {}
-
-    for i, v in ipairs(extract_table(self)) do
-        if v:IsMetadata() then
-            metadata[i] = loadstring(v.sScript)()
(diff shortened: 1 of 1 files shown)


kartoFlane:
Overall I think I lean slightly more in favor of #154 - storing metadata in a separate table would be the preferable solution if the damage list were immutable, but it is problematic when we consider that a SkillEffect's damage list can be modified after creation by mods with modApiExt's `skillBuildHook`. This means that it is possible for a mod to remove a damage instance, and if it isn't aware of metadata, we'd end up having a dangling metadata instance leftover, or mismatched metadata. Overall I think synchronization of the two lists would be a pain.

PR #154, opened 2022-08-20
Alternative method to rework how skill effect metadata is stored to fix #140

### LINKED ISSUE #140: [2.7.1] Cannot earn third Bombermechs achievement with modloader

KnightMiner:
This is an alternative approach to fixing #140 compared to #153

The third bombermech achievement relies on the second space damage in the skill effect being a killing blow, and the third skill effect targeting a bomberling. However, since the second space damage in the skill effect is a projectile, the modloader injects metadata before that effect in the final return.

To fix this, rework the skill effect so that the metadata is only stored as a new space damage if the given space damage has a script. Since the weapon causing a problem does not rely on space damage scripts, this solves the problem.

Note that currently grapples always store in a separate space damage, we could do a bit of testing to verify if we can always store them in the same space damage. That one is more difficult as the space damage is created as part of the AddGrapple function, so on the odd chance it includes a script it is too late to add the metadata.

To reduce the burden of copying skill effects (and to make modders lives easier), two new functions are added: `DamageList:AppendAll()` and `SkillEffect:AppendAll()` which copy all space damage from one skill to another.

### Breaking changes

This PR will be a breaking change for anyone currently using `SkillEffect:GetMetadata()`, `SkillEfffect:GetQueuedMetadata()`, and `DamageList:GetMetadata()` as metadata objects are now possibly not exclusively metadata. Since the size of the damage list is no longer the same, there was no way to avoid that breaking change

### DIFF
--- a/scripts/mod_loader/modapi/skills.lua
+++ b/scripts/mod_loader/modapi/skills.lua
@@ -55,7 +55,9 @@ function SpaceDamage:ToTable()
 	return result
 end
 
-function SpaceDamage:IsMetadata()
+-- returns true if this space data contains metadata
+-- it may contain other data as well, use IsMetadata to check if this is only metadata
+function SpaceDamage:HasMetadata()
     Assert.Equals("userdata", type(self), "Argument #0")
 
     if not self.sScript then
@@ -63,6 +65,36 @@ function SpaceDamage:IsMetadata()
     end
     
     return modApi:stringStartsWith(self.sScript, "return")
+end
+
+-- gets the metadata in the current space damage
+function SpaceDamage:GetMetadata()
+    Assert.Equals("userdata", type(self), "Argument #0")
+
+    if not self:HasMetadata() then
+        return {}
+    end
+
+    return loadstring(self.sScript)()
+end
+
+-- checks if this space damage is exclusively metadata
+-- note, if you plan to use GetMetadata later, it is more efficient to use
+-- local meta = spaceDamage:GetMetadata()
+-- local isMetadata = not meta.storedInSourceDamage
+function SpaceDamage:IsMetadata()
+    Assert.Equals("userdata", type(self), "Argument #0")
+
+    return self:HasMetadata() and not self:GetMetadata().storedInSourceDamage
+end
+
+function DamageList:AppendAll(list)
+    Assert.Equals("userdata", type(self), "Argument #0")
+    Assert.Equals("userdata", type(list), "Argument #1")
+
+    for i = 1, list:size() do
+        self:push_back(list:index(i))
+    end
 end
 
 function DamageList:GetMetadata()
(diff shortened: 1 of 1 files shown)


scripts/mod_loader/modapi/skills.lua
```diff
@@ -104,9 +158,10 @@ SkillEffect.AddGrapple = function(self, source, target, anim, ...)
     metadataTable.source = Point(source.x, source.y)
     metadataTable.target = Point(target.x, target.y)
     metadataTable.anim = anim
-    addDamageListMetadata(self.effect, metadataTable)
 
     oldAddGrapple(self, source, target, anim, ...)
+    -- store metadaata in the grapple space damage
+    addDamageListMetadata(self.effect, metadataTable, true)
```
kartoFlane, line 164: For grapple, we could technically create a dummy SkillEffect instance and call `oldAddGrapple` on that, then inspect the added damage instance to check if it contains a script. Then we can copy it to the actual  SkillEffect.

Though considering that the damage instance is created by the game, and not by lua, I don't think there's a high likelihood of grapple ever having a script attached to it (unless a mod overwrites it after the fact, like we do here), so I'm mentioning this more as food for though.

scripts/mod_loader/modapi/skills.lua
```diff
@@ -116,37 +171,44 @@ local function overrideProjectileOrArtillery(funcName, oldFunc)
...
-				oldFunc(self, damageInstance, projectileArt, delay)
-			else
-				oldFunc(self, damageInstance, projectileArt)
-			end
-		end
+
+        -- if the damage contains a script, add metadta first
+        if damageInstance.sScript ~= "" then
+            addDamageListMetadata(self[damageList], metadataTable, false)
+        end
```
kartoFlane, line 197: In this case, we could possibly split the script from the original damage instance, and insert it in a seprate damage instance after the projectile/artillery. I don't think there's anything technically preventing this, the only question is, will the weapon behave the same (with delay, without delay, etc)?

scripts/mod_loader/modapi/skills.lua
KnightMiner: I tested in game and it lacked a script, so I don't expect it to be a problem. It seems grapple is implemented as part of the hidden data not seen in space damage

scripts/mod_loader/modapi/skills.lua
KnightMiner: Would there be a benefit to that? We would still be changing the list size (so indexes change), and now we have a script damage that we don't know where it came from (unless the metadata references that)

kartoFlane: We would have a guarantee that projectile/artillery damage instances always contain metadata, and you don't have to fish them out by searching the list, which sounds nice on paper... But when contrasted with the fact that we change the list size anyway, and this being the more complex solution overall, I think the current solution is good.

PR #155, opened 2022-08-20
Globalize exclusiveElements in spawner_backend

Lemonymous:
### Handling of exclusive enemies have been changed in itb 1.2.78

With the new beta patch for ITB, `../scripts/spawner_backend.lua` has changed the way exclusive enemies are listed and handled; and in the process the table has been made local, where it previously was global.

This makes it difficult to update EasyEdit which relied on accessing this list.

### DIFF
--- a/scripts/mod_loader/hard-alter.lua
+++ b/scripts/mod_loader/hard-alter.lua
@@ -38,6 +38,7 @@ end
 
 globalizeLocalVariable("scripts/text.lua", "Global_Texts")
 globalizeLocalVariable("scripts/spawner_backend.lua", "WeakPawns")
+globalizeLocalVariable("scripts/spawner_backend.lua", "exclusiveElements")
 globalizeLocalVariable("scripts/game.lua", "GameObject")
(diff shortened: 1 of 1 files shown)


KnightMiner:
My only concern with this is I'd like to do a modloader release in the near future that supports the latest non-beta. Do our globalize local variable functions cause errors if the variable is missing, or will it work fine?

Lemonymous:
That is a good question, which I didn't know the answer for before.

I looked into it. The function uses regex to look for the phrase `local exclusiveElements`. spawner_backend.lua from 1.2.76b does not have this phrase, so the function will open the file, read its content, and then write the same content back into it.

Another thing to note is that the other `globalizeLocalVariable("scripts/spawner_backend.lua", "WeakPawns")` appears to do nothing now, and could possibly be removed. By viewing git blame in the itb-diff repo, we can see it was made global in 02/11/18.

This PR should not cause any harm when used with ITB 1.2.76b as far as I can tell. I have not tested it though.

KnightMiner:
I just checked and came to the same conclusion, so should be good to merge

PR #163, opened 2022-09-08
Feature failed mod popup

Lemonymous:
Depends on #145 

## New Mod Configuration Options
- **Show frienly popup when mods fail to load:** Replaces the Script Error popup with a popup showing which mods did not load, and suggestions on how to resolve it.
fail-init

- **Disable mods below user defined threshold:** Disables mods built for mod loader/game version below a user defined threshold.
modApi-config
version-popup

## Misc
- **New optional mod field `table mod.dependencies`:** if a mod has a dependency on another mod, it will fail to initialize and load if the mod it depends on either does not exist or does not successfully initialize. The reason for failing will be shown in the friendly popup.
- **Fix showButtonDialog compat bug:** Calling `sdlext.showButtonDialog` when the compatibility code did not apply, would result in always passing `nil` as the `options` argument to `sdlext.showButtonDialog`.


Lemonymous:
Should this PR draft maybe be closed and archived, since it never got finished?

I don't see myself going back to it, but it could maybe be useful if someone decides to add something similar at some point.

KnightMiner:
Yeah, if you don't have plans to finish it in the near future, closing seems reasonable. Would be helpful to mention in a comment what was not finished in case someone else looks into this.

Lemonymous:
This PR was created at a time I was also adding many other features to the mod loader, like extensions and dependencies. Due to time constraints, this PR was never merged. The current state of it is not fully known to me.

Closing it due to it being stale.

Issue #138, opened 2022-08-02
[2.7.1] Replacing the secret squad causes visual bugs


KnightMiner:
Achievements assume the original squad has achievements, however for the secret squad that is not true causing visual bugs.

In the squad selector, the medals overlap:

![image](https://user-images.githubusercontent.com/12155288/180335555-f17aefe3-e86c-4838-8146-2e37535d8405.png)

In the hangar, there are no tooltips on the achievements and the "No achievements" background is still visible:

![image](https://user-images.githubusercontent.com/12155288/180335631-b9abc25e-aced-4e96-b809-5bb36bb83111.png)

Extracted from #130

Issue #173, opened 2022-10-07
Mod loader's cached settings lags behind real data


Lemonymous:
modApi.events.onProfileChanged and modApi.events.onSettingsChanged are triggered in `modui/root.lua:79` with incorrect `Settings.last_profile`.

This likely happens because the game has not yet saved the new settings to file.

Lemonymous:
This was fixed in 5a7a953 which was included in release v2.8.0

Issue #170, opened 2022-09-23
Vanilla squads show modded victories


KnightMiner:
Since modded squads are added by replacing vanilla squads, it is possible to earn a vanilla ribbon with a modded squad. Similar to #169, there may not be anything we can do to prevent earning them, but we could at least hide the false victories.

## Approach 1: Statistics

* On profile load, scan the statistics to find the true best victory and displayed victory for each squad. Can do so by iterating all previous runs.
* If it was a vanilla squad and has the proper mechs, track that as a true best
* If it was a modded squad (improper mechs), track that as a false best
* At the end, should have true and false bests for each vanilla squad for each of the 3 lengths. If the false best is greater than the true best, save an override
* On loading the hangar with a vanilla squad, if there are any overrides display ribbons over then for the override value, use the same logic as modded squad ribbons.
* On victory,
    * If the victory is a vanilla squad, check if we can remove the override
    * If the victory is a modded squad and is greater than our previous victory, add an override for the old value.

## Approach 2: saved mod content

This approach takes just the last step from approach 1, simply tracking if we have any falsely earned victories.

* When starting the final island, store the best ribbon for that length into the modded save data
* When finishing the final island:
    * If playing a modded squad and the victory is greater than the previous, add an override for the old value.
    * If playing a vanilla squad and the victory makes the override not needed, remove it
* On loading the hangar with a vanilla squad, if there are any overrides display ribbons over then for the override value, use the same logic as modded squad ribbons.

Lemonymous:
Fixed in 9bb9f5d087181c8b4238caaf8dfb66428ebc4c45

Issue #143, opened 2022-08-03
Unfair ribbons are larger than regular ribbons, causing a visible border


KnightMiner:
![image](https://user-images.githubusercontent.com/12155288/182526243-e60900b4-65ad-43df-af85-fde84cabe7b3.png)

![image](https://user-images.githubusercontent.com/12155288/182526337-5d23885d-de8d-4de7-93e3-b8a0562343f2.png)

Unfair mode ribbons are larger, so the mod loader ribbons don't fully overlap it. Might need to draw a background or add a border to the ribbons

Background in  the squad selector would also help with #138

Lemonymous:
Fixed in 9bb9f5d087181c8b4238caaf8dfb66428ebc4c45

PR #174, opened 2022-10-07
Fix Medal/Achievement Ui

Lemonymous:
Fixes Issues #138, #143, #170, #173
Touches on Issue #176 

This PR fixes most of what was incorrect with medals/achievements/coins - See issues.

## Discrepancies
- When a mod squad with achievements replaces Secret Squad, the achievements' tooltips will be created by the mod loader and won't be located in the usual place the game would draw them.
- When a mod squad has no achievements and replaces any other squad than Secret Squad, the vanilla achievements will be covered up. However, by flicking the mouse cursor over the covered achievements, their tooltips may appear. This does not appear to be fixable.

# New functions

## Ui
The following functions has been added to facilitate creating ui code where the ui hierarchy can be seen at a glance.

### `beginUi` - self beginUi(class, ...)
Adds the instance `class` or a new ui instance of class `class` (or `Ui` if `nil`) to itself, and returns the new ui instance. Passes the arguments `...` to the ui class creator.
### `endUi` - self.parent endUi()
returns the parent ui instance. Each use of `beginU` should be paired with `endUi` when function chaining.
### `setVar` - self setVar(var, value)
Sets a variable in the ui table to the given value. Useful to set any variable in the ui instance table while function chaining.
### `format` - self format(fn, ...)
runs a function `fn(self, ...)` to format the ui instance in some way while function chaining.
### `onGameWindowResized` - onGameWindowResized()
overridable method which is called for each element attached to UiRoot when the game window is resized.

&nbsp;

### `getProfileStatistics` - table modApi:getProfileStatistics(profileName)
Returns the profile statistics for profile with name `profileName`.

### `squadIndex_game2modLoader`
Converts gameSquadIndex to modLoaderSquadIndex

### `squadIndex_modLoader2game`
Converts modLoaderSquadIndex to gameSquadIndex

### `getSquadsModSquadIndex`
Returns the modSquadIndex the squad id has, or -1 if none can be found.
Each mod squad is assigned a modSquadIndex when it is added to the mod loader, and remains constant for the duration of the application.

### `getSquadsCurrentModLoaderSquadIndex`
Returns the current modLoaderSquadIndex for the squadId, or -1 if none can be found.
Using the squad editor to change which squads are available in the hangar can change the modLoaderSquadIndex for a squad.

&nbsp;

## modApi.medals
The following tables and functions has been added to make available data required to display the true victory medals for vanilla squads.

### `correctedVanillaStats` - table
Entries in this tables have the following pattern, where the index is the squad id, and the entries have keys "2islands", "3islands" and "4islands" with string values for the corrected highest completed difficulty for this vanilla squad. It only contains corrected stats. If there is nothing to correct, then all entries will remain `nil`. This table is lazily loaded by `isVanillaScoreCorrected` and `updateVanillaRibbons`.
```lua
["Archive_A"] = {
  ["2islands"] = "EASY",
  ["4islands"] = "HARD"
},
["Rust_B"] = {
  ["3islands"] = "NONE",
  ["4islands"] = "NORMAL"
},
```

### `isVanillaScoreCorrected` - boolean isVanillaScoreCorrected(self, squadId, islands)
Returns `true` if the score for the specified vanilla squad has been corrected; otherwise `false`. This function lazily loads the table `self.correctedVanillaStats`.

### `updateVanillaRibbons` - void updateVanillaRibbons(self)
This function populates the table `self.correctedVanillaStats`.

&nbsp;

## New colors
- deco.colors.debugRed
- deco.colors.debugGreen
- deco.colors.debugBlue
- deco.colors.debugYellow
- deco.colors.debugTeal
- deco.colors.debugMagenta


scripts/mod_loader/bootstrap/constants.lua
```diff
@@ -26,4 +26,9 @@ modApi.constants = {
   PILOT_CONFIG_POD_NORMAL = 1,
   PILOT_CONFIG_POD_ADVANCED = 2,
   PILOT_CONFIG_RECRUIT = 4,
+
+  SQUAD_INDEX_START = 0,
+  SQUAD_INDEX_END = 15,
+  SQUAD_INDEX_RANDOM = 8,
+  SQUAD_INDEX_CUSTOM = 9,
```
KnightMiner, line 33: Should probably use these constants in the altered squads code as I have a couple of bound checks there for the index

Lemonymous:
PR deprecate due to most of the functionality having been added to the mod loader in some form

PR #165, opened 2022-09-18
Feature memedit

Lemonymous:
**PR Status: Ready for review/merge**

Memedit is a mod loader extension that uses `memedit.dll` to write/read directly in memory to/from addresses at some offset from a `userdata` object (like `Pawn`, `Board`, `Game`, etc).
It comes with a scanner, which will help find the correct values for each game version.
On game init (if requested by a mod), the stored addresses for the current game version will be loaded in if available.
If there are no addresses available, a recalibration process must be gone through to acquire addresses for the current game version, which then will be saved to file.

## Memedit
- Enabled by mods with extension id "memedit"
- When initialized, is located in `modApi.memedit`
- When `memedit.dll` is loaded, its output is located in `modApi.memedit.dll`
- It comes with a scanner: `modApi.memedit.scanner`
- When a mod requests memedit, and memedit is not calibrated, a calibration ui will be created, allowing the user to calibrate it
- `memedit.dll` can be loaded in debug mode with `modApi.memedit:load{ debug = true }`. When in debug mode, a memory inspect tool is available during missions, which lets the user inspect the memory of the Board, or the selected Pawn.

## Memedit function list:
- **init** - initializes memedit, and loads `memedit.dll` with addresses on file. If fully calibrated (compared vs scanner), set `memedit.calibrated` to `true`; otherwise `false`
- **isDebug** - returns `true` if dll is loaded and in debug mode; otherwise `false`.
- **load** - loads `memedit.dll` and sets it to `modApi.memedit.dll`. If fully calibrated (compared vs scanner), set `memedit.calibrated` to `true`; otherwise `false`
- **loadAddressesFromFile** - loads and returns the table of addresses we have available on file for the current game version
- **unload** - unloads any dll currently loaded, and sets `memedit.calibrated` to `false`
- **recalibrate** - starts a recalibration process, which if completed successfully will load memedit with calibrated addresses, and save those addresses to file.
- **saveAddressesToFile** - saves the input addresses to file.

## Scanner function list:
- **addToOutput** - adds the result of a scan to the output table
- **changeStatus** - used to update status internally
- **cleanupLastScan** - cleans up any current scan that might be in the middle of something
- **destroyIterator** - ends the current scanning session
- **getOutput** - returns the current output of the scanning results
- **isSuccess** - returns true if the current scanning session is successful; otherwise false.
- **logStatus** - logs out current scanning status to the console
- **reset** - resets the scanner, in preparation to start a new scanning session
- **restart** - resets and then starts a new scanning session
[...]

PR #179, opened 2022-10-15
Feature/memedit preparation

Lemonymous:
These are the changes to the mod loader in the memedit branch (feature-cutils) #165 when the memedit extension itself has been removed.

Added as a separate PR to hopefully smooth the review process for #165 

## Extension
- Adds support to have certain parts of the mod loader be "extensions" that mods can request enabled. If not requested, these "extensions" will remain dormant.

## Events
- Add onTestsuitesCreated event - This event is dispatched when after the testsuite has been created.
- Add onBoardClassInitialized - These events are dispatched when the respective object has been initialized; and can be subscribed to by mods in order to override functions within them.
- Add onPawnClassInitialized
- Add onGameClassInitialized

## Board
- Fix Board.IsTipImage and Board.IsMissionBoard
From my testing, these functions were not accurate. This change uses a well tested method that relies on checking the size of the board to know if it is a mission board or a tip image board.

## Tip Image
- Add limited capability to access the board object in the current displayed tip image. It relies on the onTipImageShown and onTipImageHidden events, which are fires for many tip images, but not all. When available, global `TipImageBoard` is set to the board of the current displayed tip image.

## Ui
- Add ORIENTATION_HORIZONTAL and ORIENTATION_VERTICAL which are constants used for which orientation a ui object will list its children.
- Add "return self" to some ui functions, so they can be used when function chaining.

## Test
- Add Tests.GetCleanTile - returns a non-pawn tile.
- Add Tests.GetNonUniqueBuildingTile - returns a non-unit tile that has not been converted to a unique building.

## Misc
- Add Assert.ShouldError
- Add memedit related globals


Lemonymous:
Pull request is deprecated. Most of the changes (Global function TipImageBoard excluded) have been either added to the mod loader, or the memedit repository, depending on where they fit the best.

Issue #188, opened 2023-02-26
No option to configure the Secret Pods in the pilot deck


KnightMiner:
The 1/23/2023 patch added a function we can override to swap out the available FTL pilots. We should make use of the pilot deck to configure those.

PR #190, opened 2023-02-26
Add FTL pilots to deck selector

tos-x:
Adds ability to set which pilots appear in secret pods.
Mod pilots with `FTL = true` will be added to the "Secret" pod deck instead of the "Pod" deck, by default.

Addresses #188


scripts/mod_loader/modapi/drops.lua
```diff
@@ -291,11 +306,13 @@ modApi.events.onModsFirstLoaded:subscribe(function()
 
 	-- copy in modded pilots
 	for _, id in ipairs(PilotListExtended) do
+	local isftl = _G[id].FTL or false
```
KnightMiner, line 309: Is this a vanilla key? If this is custom, I feel like adding an additional API on top of the legacy system just adds confusion, we could just tell modders "by default your pilot will be available in time pods, but you can use `addPilotDrop` to make it part of recruits, advanced only, or part of secret pods".

tos-x: Updated `drops.lua` with a new commit that removes the flag logic.
Need to update the wiki: https://github.com/itb-community/ITB-ModLoader/wiki/modApi#modapiaddpilotdrop

## `modApi:addPilotDrop`

| Argument name | Type | Description |
|---------------|------|-------------|
| `id` | string | ID of the pilot to add, should be the name of a global variable containing a pilot. |
| `pod` | boolean/string | Sets whether this pilot is available in time pods and as perfect island rewards. |
| `recruit` | boolean | Sets whether this pilot is available as one of the two starting recruits. |
| `ftl` | boolean | Sets whether this pilot is available in secret FTL time pods. |

Adds a pilot to be available in time pods, perfect island rewards, FTL time pods, and as a starting recruit. By default, all pilots in `Pilot_Recruits` are automatically added to the recruit deck, and all pilots in `PilotListExtended` (from `CreatePilot`) are automatically added to the time pod deck if their rarity is nonzero. By default, pilots are not added to the FTL time pod deck.

`pod` can be set to `true`, `false`, `"normal"`, or `"advanced"`. `true` makes it available in both normal and advanced edition pilot decks. `false` makes it available in neither. `"normal"` and `"advanced"` make it available in only the respective deck. This feature is not available for recruits or FTL pilots.

Example:
```lua
modApi:addPilotDrop{id = "Pilot_ID", pod = "advanced" }
modApi:addPilotDrop{id = "FTL_ID", pod = false, ftl = true }
```
&nbsp;

PR #191, opened 2023-02-27
Fix SetNeutral so it works without memEdit

tos-x:
Fix SetNeutral so it works without memEdit.


tos-x:
Verified that the tank mission works with and without memEdit enabled, with no logged warnings.

PR #193, opened 2023-02-28
Take advantage of the new injection point provided by Subset

KnightMiner:
In the most recent update, Subset added `modloader.lua` which is automatically loaded by the scripts list. This means we do not need to constantly ship an update to `scripts.lua`.

Writing this as a PR rather than a direct commit as this change would mean we only support the latest ITB version.


KnightMiner:
Note to whoever writes the changelog, tell people if they iinstalled the modloader on 1.2.88 that they need to uninstall it and verify integrity of local files before updating as we are currently shipping an outdated scripts.lua.bak.

It will probably be worth including a copy of the vanilla `scripts.lua` as an additional file in the release.

PR #200, opened 2023-04-01
Change the busy check for isAllDeployed

WolframParadoxica:
Check all three pawns at once instead of just pawn 0, to fix the issue of Deployment Effects not firing unless pawn 0 has deployed.


Lemonymous:
Thanks!

Issue #198, opened 2023-03-26
Metadata in SpaceDamage objects can pose additional challenges


Lemonymous:
### Description
SpaceDamage object's metadata is a fairly obscure thing, that is not used by many (or any?) current mod. It is not a well known thing, and could cause bugs if not handled correctly.

There exists situations where one might want to extend the script of a SpaceDamage instance in a SkillEffect that was not initially created by your mod.

In this situation, this is the order of events:
- A SkillEffect object is created
- A SpaceDamage object is created
- The SpaceDamage object is modified in for example a weapon's GetSkillEffect
- The SpaceDamage object is passed to a SkillEffect function (for example AddProjectile)
- If using `AddProjectile`, `AddQueuedProjectile`, `AddArtillery` or `AddQueuedArtillery`, the mod loader will add some metadata:
    - If the SpaceDamage object's sScript is empty, that metadata will be stored in the object's own sScript.
    - If the SpaceDamage object's sScript is not empty, that metadata will be stored in another SpaceDamage object added to the SkillEffect object.
- SkillBuild events from other mods can now have their turn in modifying the SkillEffect object and its SpaceDamage objects.

The metadata sScript starts with "return ". This causes back-concatenation to the SpaceDamage object's sScript to fail.

### Potential solutions
- The metadata could always be added to its separate SpaceDamage object.
- Function(s) could be added to handle concatenating strings to SpaceDamage objects' sScript correctly. Downside: The max length of an sScript is finite, so the metadata would take some available script space.
- Function(s) could be added to free up the metadata script, and move the metadata to a separate SpaceDamage instance.
- Other?

KnightMiner:
Separate space damage object breaks some achievements, so that route is not worth going down.

If you need to add a script to metadata, your best bet is really moving the metadata to another space damage object, or just inserting a script after the metadata. Best bet unless we ditch metadata that is

Lemonymous:
I see. I do not personally need this, but it has been posed as a question to me. And I do not have any function to redirect them to, nor know of any easy way around the issue other than just overwriting the sScript. Is that a valid solution?

What about rethinking how the metadata is connected to the SpaceDamage. In theory, you wouldn't have to store the data directly in the script. All you need is to store some unique id that can be used to fetch the data from another table.

So instead of storing the data as a table in script form, how about declaring a local variable in the script (or finding another SpaceDamage property that is inconsequential to the SpaceDamage type in question). The name of the local variable could be the unique id.

```lua
[...]

Issue #197, opened 2023-03-25
Fix SpaceDamage:HasMetadata()


tos-x:
In [skills.lua](https://github.com/itb-community/ITB-ModLoader/blob/master/scripts/mod_loader/modapi/skills.lua), `SpaceDamage:HasMetadata()` should check for "return " with a space, to be more robust.

e.g.
`return modApi:stringStartsWith(self.sScript, "return ")`
...not
`return modApi:stringStartsWith(self.sScript, "return")`

KnightMiner:
Is this causing any issues? Unless a global variable starts with the word return I cannot see this detecting a false positive.

Lemonymous:
A script calling a global function starting with "return" could yield a false positive.

The mod loader function that adds metadata always starts with "return ", so this change should still always correctly identify a SpaceDamage instance with metadata stored in its sScript.

Lemonymous:
This will become a non-issue once #198 has been fully resolved

PR #203, opened 2023-04-12
Remove SpaceDamage metadata

Lemonymous:
The data that was previously available via SpaceDamage:GetMetadata() is now available via new functions added to SpaceDamage objects when memedit is enabled:

### SpaceDamage
- GetAnimState
- GetGrappleAnim
- GetGrappleTarget
- GetMoveType
- SetAnimState
- SetGrappleAnim
- SetGrappleTarget
- SetMoveType

### DIFF
--- a/scripts/mod_loader/modapi/skills.lua
+++ b/scripts/mod_loader/modapi/skills.lua
@@ -56,39 +56,6 @@ function SpaceDamage:ToTable()
     return result
 end
 
--- returns true if this space data contains metadata
--- it may contain other data as well, use IsMetadata to check if this is only metadata
-function SpaceDamage:HasMetadata()
-    Assert.Equals("userdata", type(self), "Argument #0")
-
-    if not self.sScript then
-        return false
-    end
-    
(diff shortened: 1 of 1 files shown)


Lemonymous:
Addresses #197 and #198

KnightMiner:
Just to be clear, did we decide on removing these functions instead of having them return false/empty table? If so, looks fine to me.

Lemonymous:
I didn't feel we got 100% clarity on that, no.

I first attempted to go though the functions and have them exist, but be "inert" and return values that made sense, but it wasn't easy in every case, so I just did this instead.

I think this might actually make it easier for future maintainers, by removing what would essentially be inert dead code.

I added it as a PR though, so that we can have this discussion here. I am open to doing it differently as well.

KnightMiner:
I don't personally use the functions so I have no preference either way. I think its just tosx who mentioned using it, so just a question of whether they would rather update to not call them or keep things working.

Lemonymous:
One option could be to add the functions back in the compat file

Issue #201, opened 2023-04-09
Add hooks for Board:AddEffect and Environment:GetAttackEffect


tos-x:
Skillbuild hooks only work on tables in `_G` that have `GetSkillEffect`; SkillEffects added via other methods would also benefit from hooks or events to allow users to manipulate them.

Consider overriding `Board:AddEffect` (and possibly `Board:DamageSpace`) via [events](https://github.com/itb-community/ITB-ModLoader/wiki/events#onBoardClassInitialized).

Consider adding an environment attack hook that triggers on `GetAttackEffect` similar to how Skillbuild hooks trigger on `GetSkillEffect`.

tos-x:
From Lemonymous: For the board event, if you follow this example, and change the function; replace the logging with something like
`myEvent:dispatch(arg1, arg2, etc)`

You create the event further up with
`local myEvent = Event()`

Then subscribe to the event with
`myEvent:subscribe(someFunctWith3Args)`

KnightMiner:
Isn't skill build hook a modApiExt thing? Might make more sense to add this proposed hook there

tos-x:
Look like the skillEffect returned by `Environment:GetAttackEffect()` gets added via `Board:AddEffect()`, so there's no real need to add hooks for it, unless the same granularity (start and end hooks) is desired.

Lemonymous:
Solved in #205 and release 2.9.2

PR #205, opened 2023-04-15
Events for Board:AddEffect and Board:DamageSpace

tos-x:
Adds events that fire when `Board:AddEffect()` and `Board:DamageSpace()` are used, so their arguments can be manipulated. Addresses part of #201.

Need to update the [event wiki](https://github.com/itb-community/ITB-ModLoader/wiki/events#modapievents):

Table of contents update:
* modApi.events
	* Events related to board:
		* [onBoardAddEffect](#onBoardAddEffect)
		* [onBoardDamageSpace](#onBoardDamageSpace)

Content update:
# Events related to board

## `onBoardAddEffect`

| Argument name | Type | Description |
|---------------|------|-------------|
| `effect` | userdata | SkillEffect or SpaceDamage |

Fired just before Board:AddEffect(effect) is called, allowing the SkillEffect object to be manipulated first. The vanilla function can be called either as `Board:AddEffect(SkillEffect)` or `Board:AddEffect(SpaceDamage)`, but the event always passes a SkillEffect object. In the case of a SpaceDamage argument, a SkillEffect is created and the SpaceDamage object is added with `SkillEffect:AddDamage(SpaceDamage)`.

Example:
```lua
local handler = function(effect)
	LOG("Board:AddEffect() was used")
end

modApi.events.onBoardAddEffect:subscribe(handler)
```

&nbsp;

## `onBoardDamageSpace`

| Argument name | Type | Description |
|---------------|------|-------------|
| `spaceDamage` | userdata | SpaceDamage or Point |
| `damage` | number| Damage amount (used if argument 1 is a Point) |

Fired just before Board:DamageSpace(spaceDamage) is called, allowing the SpaceDamage object to be manipulated first. The vanilla function can be called either as `Board:DamageSpace(SpaceDamage)` or `Board:DamageSpace(Point, Int)`, but the event always passes a SpaceDamage object. In the case of point and damage arguments, `spaceDamage.loc = Point` and `spaceDamage.iDamage = damage`.

Example:
```lua
local handler = function(spaceDamage)
	LOG("Board:DamageSpace() was used")
end

modApi.events.onBoardDamageSpace:subscribe(handler)
```

&nbsp;

### DIFF
--- a/scripts/mod_loader/bootstrap/modApi.lua
+++ b/scripts/mod_loader/bootstrap/modApi.lua
@@ -128,6 +128,9 @@ t.onPawnUndeployed = Event()
 t.onPawnLanding = Event()
 t.onPawnLanded = Event()
 
+t.onBoardAddEffect = Event()
+t.onBoardDamageSpace = Event()
+
 t.onShiftToggled = Event()
 t.onAltToggled = Event()
 t.onCtrlToggled = Event()
--- a/scripts/mod_loader/modapi/board.lua
+++ b/scripts/mod_loader/modapi/board.lua
@@ -198,7 +198,24 @@ local function initializeBoardClass(board)
 
 
 	-- Override existing Board class functions here
+	BoardClass.AddEffectVanilla = board.AddEffect
+	BoardClass.AddEffect = function(self, effect)
+		Assert.Equals("userdata", type(self), "Argument #0")
+		Assert.Equals("userdata", type(effect), "Argument #1")
+		
+		modApi.events.onBoardAddEffect:dispatch(effect)
+		self:AddEffectVanilla(effect)
+	end
+	
+	BoardClass.DamageSpaceVanilla = board.DamageSpace
+	BoardClass.DamageSpace = function(self, spaceDamage)
+		Assert.Equals("userdata", type(self), "Argument #0")
+		Assert.Equals("userdata", type(spaceDamage), "Argument #1")
 
+		modApi.events.onBoardDamageSpace:dispatch(spaceDamage)
+		self:DamageSpaceVanilla(spaceDamage)
+	end
+	
 
 	modApi.events.onBoardClassInitialized:dispatch(boardClass, board)
 	modApi.events.onBoardClassInitialized:unsubscribeAll()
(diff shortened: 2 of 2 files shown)


KnightMiner:
Regarding the wiki, it might be worth adding a clarifying comment

> This event is only fired when `Board:AddEffect` is called directly from lua, it will not fire from other causes of skill effects run in internal game logic. See `###` and `###` for manipulating skill effects from weapons.

The clarification is mostly to reduce the need to understand how lua bindings work for those using the event.

Lemonymous:
Board.DamageSpace has multiple signatures. One of which it looks you have not accounted for:
```diff
- void DamageSpace(Board&,Point,int)
+ void DamageSpace(Board&,SpaceDamage)
```
Board.AddEffect has multiple signatures. Would you like both to dispatch the same event?
```diff
- void AddEffect(Board&,SpaceDamage)
+ void AddEffect(Board&,SkillEffect)
```

tos-x:
Shoot, I should have checked the function list. It would probably be
simpler to just have 2 events, I think? Then the user doesn't have to check
what the event is returning, they'll know. I'll need to find an example of
asserts for varying arguments.

KnightMiner:
#203 removed some examples of multi-signature functions

Lemonymous:
The change to Board.DamageSpace seems fine.

I added #209 which might be helpful to distinguish between when AddEffect is called with a SpaceDamage object instead of a SkillEffect object.

Lemonymous:
I am not sure, but maybe it could be appropriate to dispatch the onBoardSpaceDamage event when AddEffect is called with a SpaceDamage object?

tos-x:
Not sure I follow. I don't think there are any other overloads for `Board:AddEffect` and I didn't think we could easily tell the difference. Would prefer not to complicate it.

Lemonymous:
If you just dispatch the same event with both SpaceDamage, and SkillEffect objects, every user of the event will have to always check the passed argument.
Wouldn't it be better to make the events easy to use with a consistent argument type?

tos-x:
Aren't there tons and tons of events with "userdata" arguments where we don't worry about this? I assumed at some point it would be on the user to know their arguments. If it makes it all the way to the vanilla function (after the event) the console will give an error about the input type.

It doesn't sound like a bad idea to be able to distinguish them, but it does sound outside the scope of what I was trying to do. So I defer on how necessary it is.

tos-x:
Updated based on [GetUserdataType](https://github.com/itb-community/ITB-ModLoader/pull/209#GetUserdataType), and now handles both input options for `AddEffect`.

Lemonymous:
Thank you!

PR #207, opened 2023-04-16
Fix crash when savedata directory path contains non-ascii characters

kartoFlane:
Packaged mod loader for testing: [ITB-ModLoader-PR207.zip](https://github.com/itb-community/ITB-ModLoader/files/11240992/ITB-ModLoader-PR207.zip)

### DIFF
--- a/scripts/mod_loader/modapi/sandbox.lua
+++ b/scripts/mod_loader/modapi/sandbox.lua
@@ -13,7 +13,7 @@ function modApi:loadIntoEnv(scriptPath, envTable)
 
 	setmetatable(envTable, { __index = _G })
 	assert(pcall(setfenv(
-		assert(loadfile(scriptPath)),
+		assert(loadstring(File(scriptPath):read_to_string())),
 		envTable
 	)))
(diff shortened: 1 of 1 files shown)


Lemonymous:
This sounds like a good PR to get in.
Do you have any tips on how it could be tested? Preferably without having to change my savedata path.

kartoFlane:
The easiest way would be to simply create a lua file with non-ASCII characters in its path, and then try to load that file with `modApi:loadIntoEnv`, and verify that it doesn't crash, and that whatever the lua file returned, has been returned by this function.

To test the full case as originally reported, you'd probably need to create a new user on your PC with non-ASCII characters in its name, and run ITB as that user - as far as I understand, this should be enough to get ITB to create a separate savedata directory in that user's home directory.

I think the easier test is enough to verify this, though of course it would be nice to get confirmation with the full case, too.

Lemonymous:
Merging, as it seems to do what it says it does even if I wasn't able to test it specifically with save data for a user name with non-ascii characters.

Thank you!

PR #234, opened 2026-03-31
Adding event for store window being visible

DasKeifer:
Adding events for when the store is opened. Same changes as done in https://github.com/itb-community/ITB-ModLoader/pull/229 for additional UIs

Can be tested by adding logging hooks for the added events by pasting the code below into the console and then triggering them in game (via `store` console command)

	modApi.events.onStoreWindowShown:subscribe(function()
		LOG("STORE SHOWN!!!")
	end)
	modApi.events.onStoreWindowHidden:subscribe(function()
		LOG("STORE HIDDEN!!!")
	end)


Tatuzudo (approved):
Everything seems to be working as it should
Image

PR #235, opened 2026-03-31
SafeDamage Updated

GenericCorporatePilot2005:
added proper Crack support for AddSafeDamage and ApplyDamage, which means they'll ignore cracks to not break the tile


DasKeifer (approved):
Tested and seems to work. It shows the terrain drop icon in a weapon preview but safe damage does that for other things like forest fires as well

